PR #8, opened 2018-10-17
TArray Updates

cpfiffer:
This contains updates for the TArray type and related functions.

Highlights:
- Implements indexing interface
- Implements Iterators
- Added `similar` functions
- `TArray` is now a subtype of `AbstractArray`
- Added `display` function and a new field on the `TArray` which stores the original task to aid in printing

The `TRef` type which stores arbitrary content is outside of this request for now. I'm not quite happy yet with the implementation for the moment.

Please take a look and let me know if there's anything I've missed or additional features I can tack on here.

### DIFF
--- a/src/tarray.jl
+++ b/src/tarray.jl
@@ -23,7 +23,7 @@ for i in 1:4 ta[i] = i end  # assign
 Array(ta)                   # convert to 4-element Array{Int64,1}: [1, 2, 3, 4]
 ```
 """
-struct TArray{T,N} <: DenseArray{T,N}
+struct TArray{T,N} <: AbstractArray{T,N}
   ref :: Symbol  # object_id
   TArray{T,N}() where {T,N} = new(gensym())
 end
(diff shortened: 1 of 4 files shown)


src/tarray.jl
```diff
@@ -112,6 +134,17 @@ Base.ndims(S::TArray) = Base.ndims(task_local_storage(S.ref)[2])
 # Base.get(t::Task, S::TArray) = (t.storage[S.ref][2])
 Base.get(S::TArray) = (current_task().storage[S.ref][2])
```
mohamed82008: Can probably use this function together with `@eval` to make the forwarding definitions shorter:
```julia
for F in (:eltype, :firstindex, :lastindex, :iterate)
    @eval Base.$F(a::TArray, args...) = $F(get(a), args...)
end
```

cpfiffer: An elegant solution, thank you!

mohamed82008:
Thanks!

Issue #3, opened 2018-09-15
Refactor `TAarray`


yebai:
The `TArray` data structure implements a copy-on-write array. This is useful for task copying since standard arrays are allocated on the stack, which is only shallow copied during task copying. 

However, the current code for `TArray` has several limits: 

- type definition is over-restrictive (e.g. subtype of `DenseArray`). 
- some constructors and array-related functions are either broken or missing
- fix display for TArray

Perhaps 

- create a `TRef` type that can store value of arbitrary data types, e.g. `x = TRef([0 1  2])`

cpfiffer:
I've got most of the first and second bullet points out of the way:

- `TArray`s are now a subtype of `AbstractArray` and implement most of [the AbstractArray interface](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array-1).
- Added numerous array functions like `similar`, `iterate`, `firstindex`, `lastindex`, etc.
- I have an early-stage implementation of `TRef` that's not terrible.

But the third bullet point has proven to be a bit stickier. It appears that the reason we can't display TArrays well (particularly from the REPL) is that TArrays are only defined on a per-task basis. When the REPL executes, it appears to do so with a new task. There's no task or storage copying. 

I don't really have a good answer to how to fix this issue with how TArrays are currently implemented. Does anyone have any thoughts on a good solution here?

yebai:
> Adding a field to TArray which stores a reference to the original task that created it. display now displays that task's copy of the TArray. The prototype I have of this has performance and allocation very similar to without the field, but I don't know if it's worth carrying around a reference to a task. This way, there doesn't actually need to be any explicit task copying just to display the array. Here's an example:

Always printing using the creator's reference might be confusing to the user since `TArray` objects can be copied and subsequently mutated in other tasks, e.g. 

```julia
function f_cta()
  t = TArray(Int, 1);
  t[1] = 0;
  while true
    produce(t[1])
    t[1] = 1 + t[1]
     println(t)
  end
end

t = Task(f_cta)

consume(t) 
a = copy(t);
consume(a)
```

yebai:
Closed by #12

PR #12, opened 2018-11-02
Added warning to display(::TArray).

cpfiffer:
Regarding discussion in #3. I've only added a warning, because I think this functionality is good for REPL users to have. `show` and `println` work as intended by printing only the current task's storage.


yebai:
Thanks @cpfiffer!

PR #17, opened 2018-12-12
Drop parent field from tasks.

yebai:
This PR prepares Libtask for Julia 1.1.0, which removes a field parent from task types (see JuliaLang/julia#29483). To solve the issue, a new binary release for `Libtask` is also needed (see https://github.com/hessammehr/LibtaskBuilder2/pull/4).

### DIFF
--- a/deps/task.c
+++ b/deps/task.c
@@ -17,6 +17,10 @@ jl_task_t *jl_clone_task(jl_task_t *t)
     newt->stkbuf = NULL;
     newt->gcstack = NULL;
     JL_GC_PUSH1(&newt);
+ 
+    // Enable stack copying implementation for cloned tasks, see
+    //   https://github.com/JuliaLang/julia/pull/29578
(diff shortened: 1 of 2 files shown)


mohamed82008:
Any update here? This seems to segfault on nightly.

yebai:
This currently depends on https://github.com/hessammehr/LibtaskBuilder2/pull/4 to get through. Also, there is some change in the stack memory management mechanism for tasks in Julia 1.1. It's fixable but best be fixed after Julia 1.1 is released.

yebai:
Related issue: enable no-copy stack switching implementation 
https://github.com/JuliaLang/julia/pull/29578

yebai:
Closed in favour of https://github.com/TuringLang/Libtask.jl/pull/22.

KDr2:
I close this PR because the work it tries to do is done.

PR #22, opened 2019-02-02
Port to Julia 1.1

KDr2:
- Support multi-versions in the building system
- Port the native code, the Julia code to Julia 1.1
- Port the test cases.


yebai:
Thanks, @KDr2. Excellent work!

PR #28, opened 2019-02-15
Enable stack copying for Julia v1.1 and later

KDr2:


KDr2:
The testing coverage submitting is failed due to https://github.com/JuliaWeb/MbedTLS.jl/issues/193, many Travis users encounter this issue and they comment out the submitting as a workaround.  Here we comment out ours too.

deps/task.c
```diff
@@ -5,6 +5,19 @@
...
 #include "julia.h"
 
+jl_task_t *jl_enable_stack_copying(jl_task_t *t)
+{
+#if JULIA_VERSION_MAJOR == 1 && JULIA_VERSION_MINOR >= 1
+    if (!t->copy_stack) {
+        jl_ptls_t ptls = jl_get_ptls_states();
+        t->copy_stack = 1;
+        t->bufsz = 0;
+        memcpy(&t->ctx, &ptls->base_ctx, sizeof(t->ctx));
```
yebai: @KDr2 why do we need this line here? Isn't `t->ctx` initialised automatically when it's run?

README.md
```diff
@@ -20,7 +20,7 @@ function f_ct()
   end
 end
 
-t = Task(f_ct)
+t = Task(f_ct) |> enable_stack_copying
```
yebai: Perhaps create a sugar syntax type `CTask` or `TTask`, so that it's ok to write `t = CTask(f_ct)` ?

yebai:
Thanks, @KDr2. Looks good!

deps/task.c
KDr2:
```
  #ifdef COPY_STACKS
      if (t->copy_stack)
          memcpy(&t->ctx, &ptls->base_ctx, sizeof(t->ctx));
  #endif
```
The code above is from `task.c`, function `jl_new_task`.  According to that, if task copying is enabled for a task, its conext must be initialized with the base context. Otherwise, the context is allocated by `jl_alloc_fiber` in the function `ctx_swicth`.

README.md
KDr2: Done

deps/task.c
KDr2: Without this line, a stack copying enabled task won't have a chance to init its context. That will cause an error when it is switched to.

README.md
yebai: Thanks!

Issue #7, opened 2018-10-12
Libtask waits for ever is function body has errors.


trappmartin:
See the following minimal example:

```julia
using Libtask

function f()
    t = 0;
    while true
        t[2] = 0
        produce(t)
    end
end

t = Task(f)
consume(t)
```

trappmartin:
I tried to do that didn’t manage so far. I’ll have another try next week.

trappmartin:
@yebai could you have a look at this? I tried to resolve this issue on the branch `mt/broken_function` but with almost no success so far.

yebai:
There seems to be a bug in function `Base.schedule_and_wait`, which does not throw `Exception` as expected. Here is an MWE

```julia

function f()
           t = 0;
           while true
               t[2] = 0
               produce(t)
           end
end

Base.schedule(Task(f))
Task (failed) @0x0000000113da86d0
MethodError: no method matching setindex!(::Int64, ::Int64, ::Int64)
f() at ./REPL[4]:4

Base.schedule_and_wait(Task(f)) # never returns. 
```

mohamed82008:
`schedule(t)` passes `t` to the scheduler for execution and moves on. It doesn't throw anything:
```julia
julia> f() = error("Ignored error")
f (generic function with 1 method)

julia> g() = (schedule(Task(f)); 1)
g (generic function with 1 method)

julia> g()
1
```
`wait` will dismiss the current task until re`schedule`d or `yieldto` is called. When you do this in the main task, and no other task is going to `schedule` it again, your program will freeze. `schedule_and_wait` just calls `schedule` then `wait` so the freezing effect of `wait` is still there, when called in the main task.

yebai:
Related: https://github.com/TuringLang/Turing.jl/issues/594

PR #29, opened 2019-02-25
Catch task exception

KDr2:
To fix https://github.com/TuringLang/Libtask.jl/issues/7.

### DIFF
--- a/src/taskcopy.jl
+++ b/src/taskcopy.jl
@@ -24,14 +24,16 @@ end
 
 """
 
-task_wrapper is a wordaround for set the result/exception to the
+    task_wrapper()
+
+`task_wrapper` is a wordaround for set the result/exception to the
 correct task which maybe copied/forked from another one(the original
 one). Without this, the result/exception is always sent to the
 original task. That is done in `JULIA_PROJECT/src/task.c`, the
 function `start_task` and `finish_task`.
 
 This workaround is not the proper way to do the work it does. The
-proper way is refreshing the current_task (the variable `t`) in
+proper way is refreshing the `current_task` (the variable `t`) in
 `start_task` after the call to `jl_apply` returns.
 
 """
--- a/src/taskcopy.jl
+++ b/src/taskcopy.jl
@@ -43,18 +43,16 @@ function task_wrapper(func)
         res = func()
         ct = current_task()
         ct.result = res
-        @static if VERSION >= v"1.1.0"
-            ct.state = :done
-        end
+        isa(ct.storage, Nothing) && (ct.storage = IdDict())
+        ct.storage[:_libtask_state] = :done
         wait()
     catch ex
         ct = current_task()
         ct.exception = ex
         ct.result = ex
-        @static if VERSION >= v"1.1.0"
-            ct.state = :failed
-        end
         ct.backtrace = catch_backtrace()
+        isa(ct.storage, Nothing) && (ct.storage = IdDict())
+        ct.storage[:_libtask_state] = :failed
         wait()
     end
 end
@@ -173,10 +171,10 @@ consume(p::Task, values...) = begin
         Base.schedule(p)
         yield()
 
+        isa(p.storage, IdDict) && haskey(p.storage, :_libtask_state) &&
+            (p.state = p.storage[:_libtask_state])
+
         if p.exception != nothing
-            @static if VERSION < v"1.1.0"
-                p.state = :failed
-            end
             throw(p.exception)
         end
     end
--- a/src/taskcopy.jl
+++ b/src/taskcopy.jl
@@ -72,7 +72,10 @@ produce(v) = begin
         wait()
     end
 
-    t.state == :runnable || throw(AssertionError("producer.consumer.state == :runnable"))
+    if !(t.state in [:runnable, :queued])
+        throw(AssertionError("producer.consumer.state in [:runnable, :queued]"))
+    end
+    if t.state == :queued yield() end
     if empty
         Base.schedule_and_wait(t, v)
         ct = current_task() # When a task is copied, ct should be updated to new task ID.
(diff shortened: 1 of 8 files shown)


src/taskcopy.jl
```diff
@@ -22,7 +22,42 @@ function enable_stack_copying(t::Task)
...
+function task_wrapper(func)
+    () ->
+    try
+        res = func()
+        ct = current_task()
+        ct.result = res
+        @static if VERSION >= v"1.1.0"
+            ct.state = :done
+        end
+        wait()
```
yebai: Any reason why we need `wait` here since the task is either completed successfully or already failed?

src/taskcopy.jl
```diff
@@ -22,7 +22,42 @@ function enable_stack_copying(t::Task)
...
+    catch ex
+        ct = current_task()
+        ct.exception = ex
+        ct.result = ex
+        @static if VERSION >= v"1.1.0"
+            ct.state = :failed
+        end
+        ct.backtrace = catch_backtrace()
+        wait()
```
yebai: See above.

src/taskcopy.jl
```diff
@@ -22,7 +22,42 @@ function enable_stack_copying(t::Task)
...
+`start_task` after the call to `jl_apply` returns.
+
+"""
+function task_wrapper(func)
+    () ->
+    try
+        res = func()
+        ct = current_task()
+        ct.result = res
+        @static if VERSION >= v"1.1.0"
```
yebai: Silly question, what happens when using Julia 1.0?

src/taskcopy.jl
```diff
@@ -129,5 +167,16 @@ consume(p::Task, values...) = begin
         push!(p.storage[:consumers].waitq, ct)
     end
 
-    p.state == :runnable ? Base.schedule_and_wait(p) : wait() # don't attempt to queue it twice
+    if p.state == :runnable
+        Base.schedule(p)
+        yield()
+
+        if p.exception != nothing
+            @static if VERSION < v"1.1.0"
```
yebai: To confirm, you want to use `VERSION < v"1.1.0"`, instead of  `VERSION >= v"1.1.0"`?  Since all above examples use ` VERSION >= v"1.1.0"`.

yebai:
Thanks, @KDr2.  I have tested the MWE from https://github.com/TuringLang/Turing.jl/issues/586. `Libtask` can successfully catch the exception/error now. Below is an example output

```julia

julia> sample(mf, SMC(10))
ERROR: BoundsError: attempt to access 1-element Array{Int64,1} at index [2]
Stacktrace:
 [1] consume(::Task) at /Users/hg344/.julia/packages/Libtask/ldg9e/src/taskcopy.jl:178
 [2] consume at /Users/hg344/.julia/packages/Turing/sghu6/src/core/container.jl:35 [inlined]
 [3] consume(::Turing.Core.ParticleContainer{Turing.Core.Trace,Turing.Model{Tuple{:x},Tuple{},getfield(Main, Symbol("###inner_function#369#5")){Array{Int64,1}},NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}}}}) at /Users/hg344/.julia/packages/Turing/sghu6/src/core/container.jl:151
 [4] sample(::Turing.Model{Tuple{:x},Tuple{},getfield(Main, Symbol("###inner_function#369#5")){Array{Int64,1}},NamedTuple{(),Tuple{}},NamedTuple{(),Tuple{}}}, ::SMC{Union{},typeof(Turing.Utilities.resample_systematic)}) at /Users/hg344/.julia/packages/Turing/sghu6/src/inference/smc.jl:81
 [5] top-level scope at none:0

```

src/taskcopy.jl
KDr2: This wrapper function is for copied tasks. If we don't copy task, only to catch exceptions occur in a task from the `consume` function, this wrapper function can be deleted.

When a task is copied before it is started, all things go well too.

But if we copy a started task, the copies and the original task are all in the call to `jl_apply`, see https://github.com/JuliaLang/julia/blob/master/src/task.c#L572. If a copy returns from that function(the copy is successfully done, or failed by exceptions), the original task will be marked done or failed by calling to `finish_task` on `t` which is the original task. We can modify this function to refresh t by `t = ptls->current_task;` after `jl_apply` returns and before calling to `finish_task`, then t will become to the copy task.

But we don't want to modify and re-compile Julia, so I make this wrapper. In this wrapper, we catch the exception or get the result, then set them to the real current task, the copy task, to prevent Julia from sending them to the original one. Also, the `wait()` at the end is to prevent this current copy task from finished -- if it finishes, its original task will be marked as done incorrectly by `src/task.c`.

src/taskcopy.jl
KDr2: It's for the same purpose as above.

src/taskcopy.jl
KDr2: While using Julia 1.0, if we set the state of a copied task to done or failed in this wrapper, a segment fault occurs. It seems to be caused by GC.  I didn't investigate deep enough here.

src/taskcopy.jl
KDr2: As mentioned above, I set the task state in the wrapper only when `VERSION >= v"1.1.0"`. So here when an exception is detected, I set the task state as complementation when Julia v1.0 is used.

src/taskcopy.jl
KDr2: When a copied task finishes (successes or fails), its original task instead of itself will be finished with `finish_task`, this is a bug of Libtask we don't know before. This PR fixes it in a not-beautiful way.

If we want to fix this properly, we should change the code of Julia, and the change should be backported. It will be a long way I think.

src/taskcopy.jl
yebai: Julia GC has an early-GC mechanism for tasks: once a task is finished, its stack will be freed immediately. This mechanism might be the cause for segfault here since a running task tries to label *itself* as `done`/`failed`. Maybe Julia 1.1 has removed/changed this early-GC mechanism, but in general, it's strange that a task tries to changes its own state. How about:

- add an entry to `ct.storage`, e.g. `ct.storage[:state] = ...`
- delay this operation to a later stage, e.g. in the `consume` function? 

This might be safer and would allow you to unify the code of changing task states for Julia 1.1 and 1.0.

KDr2: Done.

PR #35, opened 2019-06-03
Extract stacktrace from the failed task

KDr2:


yebai:
Nice improvements, many thanks @KDr2!

PR #42, opened 2019-09-05
[WIP] Resume from task without `produce`

### LINKED ISSUE #41: `consume` never returns if task function doesn't contain `produce`
The following code would halt

```julia

using Libtask

function f_ct()
   t = 0;
   while t<3
     println(t)
     t = 1 + t
   end
   return t
end

t = CTask(f_ct)

julia> consume(t) # Never returns. 
0
1
2

```

This is likely due to the `consumer` task never gets waken up if the `producer` task exit before any `produce` statement. Maybe a better behaviour for this kind of tasks is to wake up the `producer` task with the `producer` task's `return` value?

KDr2:
Try to fix #41.

Status Quo: Still has problems when tasks throw exceptions.

### DIFF
--- a/src/taskcopy.jl
+++ b/src/taskcopy.jl
@@ -42,21 +42,24 @@ proper way is refreshing the `current_task` (the variable `t`) in
 
 """
 function task_wrapper(func)
-    () ->
-    try
+    () -> begin
         ct = current_task()
-        res = func()
-        ct.result = res
-        isa(ct.storage, Nothing) && (ct.storage = IdDict())
-        ct.storage[:_libtask_state] = :done
-        wait()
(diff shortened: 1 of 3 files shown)


yebai:
Hi @KDr2, the current `produce`/`consume` implementation in `Libtask` is taken from early versions of Julia. These functions have been now replaced by `Channel` after Julia 1.0. Maybe we should consider re-implementing `produce`/`consume` by wrapping around `Channel` API? This would make `Libtask` more in line with the Julia std library.

KDr2:
> Maybe we can consider re-implementing `produce`/`consume` by wrapping around `Channel` API

I will try this :)

Issue #55, opened 2020-03-29
Test errors on DynamicPPL due to Libtask


devmotion:
CI tests of DynamicPPL (see https://travis-ci.com/github/TuringLang/DynamicPPL.jl/jobs/306791553 and https://github.com/TuringLang/DynamicPPL.jl/pull/52) fail on Julia 1.4 and Julia master due to the error
```julia
ERROR: LoadError: The following package names could not be resolved:
 * Libtask (not found in project or manifest)
Stacktrace:
 [1] pkgerror(::String) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Types.jl:53
 [2] ensure_resolved(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; registry::Bool) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Types.jl:794
 [3] ensure_resolved at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Types.jl:759 [inlined]
 [4] build(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; verbose::Bool, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:697
 [5] build at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:682 [inlined]
 [6] #build#108 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:680 [inlined]
 [7] build at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:680 [inlined]
 [8] #build#105 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:677 [inlined]
 [9] build(::String) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:677
 [10] top-level scope at /home/travis/.julia/packages/Libtask/K7OAU/src/Libtask.jl:7
 [11] include(::Module, ::String) at ./Base.jl:377
 [12] top-level scope at none:2
[...]

PR #56, opened 2020-03-29
Provide error message instead of building Libtask

devmotion:
This PR addresses https://github.com/TuringLang/Libtask.jl/issues/55.

I'm not sure if that's the best way to solve it but it avoids calling `Pkg.build("Libtask")` which apparently fails in CI tests. Moreover, it fixes the typo due to which, as far as I can see, the package was rebuilt each time it was loaded.

### DIFF
--- a/.github/workflows/Testing.yaml
+++ b/.github/workflows/Testing.yaml
@@ -19,7 +19,6 @@ jobs:
           - '1.2'
           - '1.3'
           - '1.4'
-          - 'nightly'
         os:
           - ubuntu-latest
           - windows-latest
--- a/.github/workflows/Testing.yaml
+++ b/.github/workflows/Testing.yaml
@@ -7,23 +7,35 @@ on:
     # tags: '*'
     release:
   pull_request:
-    types: [opened, synchronize, reopened]
 
 jobs:
   test:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ['1.0.5', '1.1.1', '1.2.0', '1.3.0', '1.4.0']
-        julia-arch: [x64, x86]
-        os: [ubuntu-latest, windows-latest, macOS-latest]
+        version:
+          - '1.0'
+          - '1.1'
+          - '1.2'
+          - '1.3'
+          - '1.4'
+          - 'nightly'
+        os:
+          - ubuntu-latest
+          - windows-latest
+          - macOS-latest
+        arch:
+          - x64
+          - x86
         exclude:
           - os: macOS-latest
-            julia-arch: x86
+            arch: x86
 
     steps:
-      - uses: actions/checkout@v1.0.0
-      - uses: julia-actions/setup-julia@latest
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
         with:
-          version: ${{ matrix.julia-version }}
-      - uses: julia-actions/julia-runtest@master
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: julia-actions/julia-buildpkg@latest
+      - uses: julia-actions/julia-runtest@latest
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 license = "MIT"
 desc = "C shim for task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
-version = "0.3.3"
+version = "0.3.4"
 
 [deps]
 BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,5 +1,6 @@
 function find_prev_tag(tag)
     project_root = (@__DIR__) |> dirname |> abspath
+    run(`git -C $project_root fetch --tags`)
     tags = readlines(`git -C $project_root tag`)
     sort!(tags)
     idx = indexin([tag], tags)[1]
--- a/src/Libtask.jl
+++ b/src/Libtask.jl
@@ -2,12 +2,14 @@ module Libtask
 
 export CTask, consume, produce, TArray, get, tzeros, tfill, TRef
 
-if !isfile(joinpath((@__DIR__) |> dirname, "deps/desp.jl"))
-    import Pkg
-    Pkg.build("Libtask")
+# Try to load the binary dependency
+if isfile(joinpath(@__DIR__, ".." , "deps", "deps.jl"))
+    include("../deps/deps.jl")
+    check_deps()
+else
+    error("Libtask is not properly installed. Please run `import Pkg; Pkg.build(\"Libtask\")`")
 end
 
-include("../deps/deps.jl"); check_deps();
 include("taskcopy.jl")
 include("tarray.jl")
 


devmotion:
I added a missing build step to the CI tests and disabled tests on Julia master again since they were not finished after > 30 minutes.

devmotion:
As expected with the current setup, tests fail after increasing the version number. Wouldn't it be possible to decouple the version of the binary dependency and the version of the Julia package, since tests passed, e.g., with the binary dependency of version 0.3.3?

devmotion:
IMO, for Julia >= 1.3 we should use [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil) if possible. However, for the time being, since we still want to support Julia < 1.3, I guess the best way would be to build the binary dependencies in a separate `LibtaskBuilder` repository, in the same way as, e.g., [SundialsBuilder](https://github.com/SciML/SundialsBuilder) builds the binaries for [Sundials.jl](https://github.com/SciML/Sundials.jl). In that way building the binaries would be decoupled from the Julia package, and hence we wouldn't get test errors in non-breaking updates of the Julia package.

KDr2:
> IMO, for Julia >= 1.3 we should use [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil) if possible. However, for the time being, since we still want to support Julia < 1.3, I guess the best way would be to build the binary dependencies in a separate `LibtaskBuilder` repository, in the same way as, e.g., [SundialsBuilder](https://github.com/SciML/SundialsBuilder) builds the binaries for [Sundials.jl](https://github.com/SciML/Sundials.jl). In that way building the binaries would be decoupled from the Julia package, and hence we wouldn't get test errors in non-breaking updates of the Julia package.

Yeah, the current binary-providing process has flaws when we bump up the version in Project.toml but not release it. But separating it to another repo pose maintenance burden, and considering common users should not use an unreleased version, I added a retrying mechanism to use the binaries for the previous version. That re-try mechanism became nugatory due to the shallow clone of repo, I fixed it with a new commit on this branch.

devmotion:
> Yeah, the current binary-providing process has flaws when we bump up the version in Project.toml but not release it. But separating it to another repo pose maintenance burden, and considering common users should not use an unreleased version, I added a retrying mechanism to use the binaries for the previous version.

Actually I think the maintenance burden would be much lower since the binaries would have to be rebuilt only occasionally but not for every update of the Julia package. For instance, the last proper commit (i.e., excluding changes of CITATION.bib and CompatHelper) in SundialsBuilder dates back to [July 2018](https://github.com/SciML/SundialsBuilder/commits/master) whereas Sundials.jl was updated and fixed [many times since July 2018](https://github.com/SciML/Sundials.jl/commits/master).

The current setup already caused some additional problems again since it was not possible to release the latest commit on the master branch with JuliaRegistrator (it failed because a tag for the same version already exists for the commit in which the package version was updated) and even releasing that commit caused test errors in the registry: https://github.com/JuliaRegistries/General/pull/11820

yebai:
Maybe add a parameter to specify a previous version of binary as a fallback? For example, on top of the version information (say `v0.21`) in `Project.toml` , we could specify a parameter in `deps/build.jl` to determine a fallback binary release (say `v0.10`).  If the current version `v0.21` is not released on Github yet, we could simply download the binary from `v0.10`. This way, we only need to occasionally update this version parameter in `deps/build.jl` to a newer released version and avoid maintaining two repos.

KDr2:
> Maybe add a parameter to specify a previous version of binary as a fallback? For example, on top of the version information (say `v0.21`) in `Project.toml` , we could specify a parameter in `deps/build.jl` to determine a fallback binary release (say `v0.10`). If the current version `v0.21` is not released on Github yet, we could simply download the binary from `v0.10`. This way, we only need to occasionally update this version parameter in `deps/build.jl` to a newer released version and avoid maintaining two repos.

Hmm, actually we are now already doing so, but instead of specifying a fallback version, we find the last release version to use. These two methods have the same effect.

The issue is, when you can't find a compatible version to fallback, the release will fail. This is what @devmotion pointed out.

How about this:
1. create a separate repo for the binary releasing, but this repo contains nothing rather than github actions to clone the Libtask repo, build, and upload the library
2. So, we can specify the certain dylib to download in Libtask's build.jl
3. When the code of dylib is updated, we make a new release on the separate repo, and the new dylib will be built and uploaded to the separate repo
4. update Libtask's build.jl to use the latest dylib URL, then tag and release Libtask.

In step 3, we can write a script to sync the tags between the two repos, so that we don't need to tag on the separate repo manually. We tag on Libtask, the script tag the separate repo and make a release, then build and upload the dylib, then we release Libtask.

KDr2:
Another way to fix the release issue is to build the dylib locally if no binary found for the current release. ThArrays uses this approach.

devmotion:
> create a separate repo for the binary releasing, but this repo contains nothing rather than github actions to clone the Libtask repo, build, and upload the library
> So, we can specify the certain dylib to download in Libtask's build.jl
> When the code of dylib is updated, we make a new release on the separate repo, and the new dylib will be built and uploaded to the separate repo
> update Libtask's build.jl to use the latest dylib URL, then tag and release Libtask.

Isn't that basically what BinaryBuilder suggests to do and what is done in the SundialsBuilder + Sundials.jl example above in a more clearly separate way? I don't really see how keeping the Libtask sources for the binaries in Libtask.jl and then cloning the repo from the other repo would be easier to handle then just keeping the binary-related things in one repo. One could still just make a new release of the binary if needed, without having to clone (and possibly to adjust) the Julia repo, and then update the build.jl accordingly to point to the new release.

IMO currently the process of building the binaries and making a new release of Libtask.jl is too intertwined, and hence seems to fail quite regularly. I guess, separating both steps by using dedicated repos could be helpful.

KDr2:
> IMO, for Julia >= 1.3 we should use Yggdrasil if possible. However, for the time being, since we still want to support Julia < 1.3, I guess the best way would be to build the binary dependencies in a separate LibtaskBuilder repository

Hmm... I think this limitation doesn't apply to us, because in our build scripts( see [here](https://github.com/TuringLang/Libtask.jl/blob/master/deps/build_dylib.sh) and [here](https://github.com/TuringLang/Libtask.jl/blob/master/deps/build_tarballs.jl) ) we download different versions of Julia then build the binary. So we can simply submit our build script to Yggdrasil to generate a `*_jll.jl` package which Libtask can depend on.

devmotion:
AFAIK the problem is not the build process but that Yggdrasil uses the new artifacts system that was only added to Pkg in Julia 1.3 (https://julialang.github.io/Pkg.jl/dev/artifacts/). So the main issue is that in Julia < 1.3 only Julia packages can be installed and added as dependencies.

KDr2:
Oh sorry for the misunderstanding, but this doesn't matter either. What we are doing now is downloading the tarball (i.e. artifacts) from Libtask's GitHub Release page, we can just change the download URL to the `*_jll.jl` package's, rather than use the artifacts mechanism.

devmotion:
The officially recommended way to deal with this problem if one wants to support Julia < 1.3 is to generate a build.jl file from the Yggdrasil release: https://github.com/JuliaPackaging/Yggdrasil#binaryproviderjl

KDr2:
Yeah, we now have and use this build file too, there’s a build.jl for each release along with the binary tarballs on the release page.

KDr2:
PR to Yggdrasil filed: https://github.com/JuliaPackaging/Yggdrasil/pull/699

devmotion:
Great, hopefully that will make the release process smoother!

Issue #53, opened 2020-03-11
Improving artefacts uploading process 


yebai:
Maybe we can modify the [build](https://github.com/TuringLang/Libtask.jl/blob/master/.github/workflows/BuildDylib.yaml) script of `Libtask`, so that a tag is automatically created if the tag does not exist? This would allow uploading artefacts for new Libtask versions even before a new version is accepted by JuliaRegistry.

devmotion:
IMO the correct approach would be to separate building and releasing the binaries into a separate package (and finally to Yggdrasil at some point), as outlined in https://github.com/TuringLang/Libtask.jl/pull/56#issuecomment-605689423.

devmotion:
BTW I just noticed that the current setup with automatically creating tags implies that only exactly the commit for which the tag was created can be registered in the registry (e.g., if the version number is updated in the PR, then a tag is created for this commit and it is impossible to use JuliaRegistrator on the master branch or the merge commit afterwards).

KDr2:
Close in favor of #58

Issue #57, opened 2020-03-31
Automerge tests in the general registry fail


devmotion:
I retriggered the [PR to the general registry](https://github.com/JuliaRegistries/General/pull/11820) on the master branch, i.e., the updated PR includes also the fix to `deps/build.jl`. However, the automerge tests still fail due to the following error:
```julia
┌ Error: Error building `Libtask`: 
│ fatal: not a git repository (or any of the parent directories): .git
│ ┌ Error: Download failed: curl: (22) The requested URL returned error: 404 Not Found
│ └ @ Base download.jl:43
│ ERROR: LoadError: failed process: Process(`git -C /tmp/jl_NXjesH/packages/Libtask/kykuJ fetch --tags`, ProcessExited(128)) [128]
│ 
│ Stacktrace:
│  [1] pipeline_error at ./process.jl:525 [inlined]
│  [2] run(::Cmd; wait::Bool) at ./process.jl:440
│  [3] run at ./process.jl:438 [inlined]
│  [4] find_prev_tag(::String) at /tmp/jl_NXjesH/packages/Libtask/kykuJ/deps/build.jl:3
│  [5] include_build_script(::SubString{String}, ::Bool) at /tmp/jl_NXjesH/packages/Libtask/kykuJ/deps/build.jl:45
│  [6] top-level scope at /tmp/jl_NXjesH/packages/Libtask/kykuJ/deps/build.jl:65
│  [7] include(::String) at ./client.jl:439
│  [8] top-level scope at none:5
│ in expression starting at /tmp/jl_NXjesH/packages/Libtask/kykuJ/deps/build.jl:65
└ @ Pkg.Operations /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:892
```
It seems it's not possible to run https://github.com/TuringLang/Libtask.jl/blob/4e201c561cd73c2fc22ccff854f5865ef4b06cc9/deps/build.jl#L3 in the registry CI since the project directory is not downloaded as a git repository there.

I guess it would be more appropriate to fix a specific version but that seems impossible in the current setup where the latest version might not exist during testing?

devmotion:
I just updated the PR with the latest commit on the master branch. As expected, the problem was fixed by https://github.com/TuringLang/Libtask.jl/pull/58.

PR #58, opened 2020-04-02
Use Yggdrasil to host binary

KDr2:


KDr2:
Will fix #53 and #57.

devmotion:
I don't think this is the ideal solution yet since we still have to keep the `build_dylib.sh` and the `build_tarballs.jl` files around (which BTW are different from the ones that are checked in Yggdrasil - so they have to be synced manually) and have some rudimentary build CI tests to make sure that changes to the build files don't break anything. Maybe `libtask.c` and `Makefile` should be moved to Yggdrasil (and maybe `Makefile` even just integrated in the `build_tarballs.jl` script)? But I guess we can keep iterating until we find a better solution.

In general, I think we could just remove the commented out or unneeded parts since they are still available in the git history anyways. In the build Github action the tagging part is not needed either anymore I guess? At least I think it caused problems in the last PR. I.e., I would suggest removing at least `deps/gh-auto-tag` and `deps/gh-update-to-release`.

Maybe we could also switch from building binaries in every PR to just building binaries if one of the files `deps/Makefile`, `deps/libtask.c`, or `deps/build_tarballs.jl` changed using https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-paths (we do not have to build binaries if only `deps/build.jl` changed, so I specified the files explicitly):
```julia
on:
  push:
    paths:
    - 'deps/Makefile'
    - 'deps/libtask.c'
    - 'deps/build_tarballs.jl'
    - 'deps/build_dylib.sh'
```
The Julia integration, however, won't be tested (Libtask.jl will still use the binaries served from Libtask_jll), so maybe it would actually make sense to move these files to Yggdrasil. Additionally, maybe the build CI should run with a newer Julia version?

I also ran the script to generate the `build.jl` file, as suggested in [the docs of Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil#binaryproviderjl) (with Julia 1.3). Initially it failed due to the use of `@__FILE__` in the `build_tarballs.jl` file in Yggdrasil. I think the script in Yggdrasil should be updated and `build_dylib.sh` should be added to the list of sources. After fixing that it worked for me, so IMO we don't need the script for retrieving the hashes (that's done automatically by the script in Yggdrasil) and we could rather stick to the generated `build.jl` and add an additional function for filtering the current version.

devmotion:
BTW the logic for picking the correct version depending on the Julia version could also be moved to `src/Libtask.jl`. I guess that might be needed anyways when moving to the artifacts system at some point, since then loading `using Libtask_jll` would enable the use of all libtask libraries? I'm also wondering, is the current approach of filtering based on the Julia version problematic when switching between different Julia versions, i.e., if `deps/deps.jl` was generated with Julia 1.3, is it updated when Libtask is loaded with Julia 1.2 the next time?

KDr2:
You are right, I did some update of this PR:
- move the version selection to deps.jl
- remove unused files, update the DylibBuild workflow to make it only run on relevant files updated

About the artifacts system and `using *_jll`, I think maybe we could not adopt it before giving up the supports of Julia 1.2 and earlier, so let's use the build.jl and deps.jl for the time beings?

About where to place the source and build scripts (task.c Makefile, build_tarballs.jl, dylib_build.sh), I didn't find an ideal way either. No one puts code source to Yggdrasil, and a separate repo for it is also too heavy because we don't need one to release and host the binary lib now.

KDr2:
About generating `build.jl` using the script provided by Yggdrasil, I can't run it either,
On Julia 1.5:
``` sh
$ julia generate_buildjl.jl L/Libtask/build_tarballs.jl
[ Info: Build tarballs script: L/Libtask/build_tarballs.jl
[ Info: Repo name: JuliaBinaryWrappers/Libtask_jll.jl
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `https://github.com/JuliaRegistries/General.git`
ERROR: LoadError: MethodError: no method matching registered_paths(::Pkg.Types.EnvCache, ::Base.UUID)
Closest candidates are:
  registered_paths(::Pkg.Types.Context, ::Base.UUID) at /home/kdr2/Work/julia/julia/usr/share/julia/stdlib/v1.5/Pkg/src/Types.jl:1190
Stacktrace:
 [1] top-level scope at /home/kdr2/Work/julia/Yggdrasil/generate_buildjl.jl:164
 [2] include(::Function, ::Module, ::String) at ./Base.jl:381
 [3] include(::Module, ::String) at ./Base.jl:369
 [4] exec_options(::Base.JLOptions) at ./client.jl:289
 [5] _start() at ./client.jl:491
in expression starting at /home/kdr2/Work/julia/Yggdrasil/generate_buildjl.jl:157
```
On Julia 1.3:
```
l$ julia generate_buildjl.jl L/Libtask/build_tarballs.jl
ERROR: LoadError: UndefVarError: BinaryPlatforms not defined
Stacktrace:
 [1] include at ./boot.jl:328 [inlined]
 [2] include_relative(::Module, ::String) at ./loading.jl:1094
 [3] include(::Module, ::String) at ./Base.jl:31
 [4] top-level scope at none:2
 [5] eval at ./boot.jl:330 [inlined]
 [6] eval(::Expr) at ./client.jl:433
 [7] top-level scope at ./none:3
in expression starting at /home/kdr2/.julia/packages/BinaryBuilder/o9Hrl/src/BinaryBuilder.jl:6
ERROR: LoadError: Failed to precompile BinaryBuilder [12aac903-9f7c-5d81-afc2-d9565ea332ae] to /home/kdr2/.julia/compiled/v1.3/BinaryBuilder/WHzgM.ji.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] compilecache(::Base.PkgId, ::String) at ./loading.jl:1253
 [3] _require(::Base.PkgId) at ./loading.jl:1013
 [4] require(::Base.PkgId) at ./loading.jl:911
 [5] require(::Module, ::Symbol) at ./loading.jl:906
 [6] include at ./boot.jl:328 [inlined]
 [7] include_relative(::Module, ::String) at ./loading.jl:1094
 [8] include(::Module, ::String) at ./Base.jl:31
 [9] exec_options(::Base.JLOptions) at ./client.jl:295
 [10] _start() at ./client.jl:468
in expression starting at /home/kdr2/Work/julia/Yggdrasil/generate_buildjl.jl:3
```
But we I think we should modify the generate `build.jl` even if that script runs successfully, unless we figure out a better way to deal with the version selection of the libs.

devmotion:
> About generating build.jl using the script provided by Yggdrasil, I can't run it either

It worked on Julia 1.3 for me (after fixing the problem with `@__FILE__`, as I mentioned). Maybe you have to update your environment or some packages on Julia 1.3? I used a local environment and just added BinraryBuilder.

.github/workflows/TagBot.yml
```diff
@@ -1,11 +0,0 @@
-name: TagBot
-on:
-  schedule:
-    - cron: 0 * * * *
```
devmotion, line 4: You still want to keep TagBot.yml since that will tag a new release when the PRs to the registry are merged. Maybe just change it to once per day (I think it's not needed to run it hourly):
```yml
     - cron: '0 0 * * *'
```

KDr2: Sorry, not intentionally to delete this, I confused it with the auto-tag job, updated in a new commit.

devmotion:
> You are right, I did some updates of this PR:
>
>    move the version selection to deps.jl
>    remove unused files, update the DylibBuild workflow to make it only run on relevant files updated
>
> About the artifacts system and using *_jll, I think maybe we could not adopt it before giving up the supports of Julia 1.2 and earlier, so let's use the build.jl and deps.jl for the time beings?

Sure, for the time being I guess it's easiest if we keep using the `buid.jl` and `deps.jl` files instead of hacking something together.

However, I think the version selection should not be moved to `deps.jl` but rather to `src/Libtask.jl`. IMO that has some advantages:
- We can just generate the standard `deps/build.jl` file and so update it easily if there is a new release of Libtask_jll.
- If we don't use it yet, the current setup in `deps/build.jl` won't work with Libtask_jll whereas the other approach does.

So I think we should just add the following to src/Libtask.jl after loading the `deps/deps.jl` file (basically moving from taskcopy.jl):
```julia
@static VERSION >= v"1.4"
	const libtask = libtask_1_4
elseif VERSION >= v"1.3"
	const libtask = libtask_1_3
elseif VERSION >= v"1.2"
	const libtask = libtask_1_2
elseif VERSION >= v"1.1"
	const libtask = libtask_1_1
else
	const libtask = libtask_1_0
end
```

devmotion:
> About where to place the source and build scripts (task.c Makefile, build_tarballs.jl, dylib_build.sh), I didn't find an ideal way either. No one puts code source to Yggdrasil, and a separate repo for it is also too heavy because we don't need one to release and host the binary lib now.

Yes, most packages download source code from some external repository but it's very common to include patches in the subdirectory `./bundled` in Yggdrasil. And some packages even include source files, e.g., https://github.com/JuliaPackaging/Yggdrasil/blob/c9d46c4474502496a9933bf577bb5a69b8caf94e/S/SuiteSparse/bundled/SuiteSparse_wrapper/SuiteSparse_wrapper.c. Typically the build_dylib.sh is just written out explicitly in `build_tarballs.jl` when defining the `script` variable. The Makefile is often part of the external source code, though, but it just seems to set flags for the compiler which is typically also done when defining the `script` variable. So maybe both `build_dylib.sh` and the `Makefile` could actually be removed.

KDr2:
It has an issue putting the version selection to `Libtask.jl` and keep the generate `deps.jl` untouched:
In `deps.jl`, it will try to load every lib file and test them (libtask_v1_0, *v1_1...) which may cause errors, e.g., when Julia 1.3 is used, it will fail to load libtask_v1_0 (API not compatible, what is the reason we need multi-version libs). So we must update the `deps.jl` file to avoid this.

KDr2:
Now, I managed to generate `build.jl` by running `julia generate_buildjl.jl L/Libtask/build_tarballs.jl`, then make this commit: https://github.com/TuringLang/Libtask.jl/pull/58/commits/bd0687796807e83c3f9a8b0d8567eafcab6040d6, but found this generated `build.jl` is not compatible with some version of Julia, see 
https://github.com/TuringLang/Libtask.jl/pull/58/checks?check_run_id=554626983#step:4:34

So I decided to revert the commit and use my version. now:

- build and host binary within Yggdrasil
- use a (manual update but need not update frequently) build.jl to generate deps.jl. 
- select lib version in `deps.jl` to avoid error on version mismatch

Is this OK?

devmotion:
>  In deps.jl, it will try to load every lib file and test them (libtask_v1_0, *v1_1...) which may cause errors, e.g., when Julia 1.3 is used, it will fail to load libtask_v1_0 (API not compatible, what is the reason we need multi-version libs). So we must update the deps.jl file to avoid this.

Isn't the only check loading the libraries? Surely the incompatible libraries can't be used but would it even fail to open the libraries?

The third reason for why I thought it might be preferable to not handle the selection in the `Pkg.build` step is that I'm not sure if packages are rebuilt automatically if you switch Julia versions (since I guess usually binaries only depend on the OS). Do you know if this is the case? I think it would be a bit unfortunate if everyone has to rebuild Libtask manually everytime one switches Julia versions, even if it's e.g. just a dependency of Turing.

KDr2:
> Isn't the only check loading the libraries? Surely the incompatible libraries can't be used but would it even fail to open the libraries?

Yeah, it checks if every lib file exists and try to load them by calling `dlopen`, `dlopen` throws errors when symbols can't be resolved, so I used a product filter to modify the generated build.jl(https://github.com/TuringLang/Libtask.jl/blob/master/deps/build.jl#L12) to eliminate the incompatible libs.

devmotion:
The errors are unrelated and explained in the [Yggdrasil README](https://github.com/JuliaPackaging/Yggdrasil#binaryproviderjl):
> Note: you have to manually add prefix as the first argument to all Product constructors in the generated build.jl files. This is necessary because the syntax between BinaryBuilder v0.2+ and BinaryProvider has diverged.

After adding `prefix` to the `LibraryProduct` calls (i.e., changing it to `LibraryProduct(prefix, ....)`), I am able to build and load Libtask with Julia 1.0, 1.1, 1.2, 1.3, and 1.4..

.github/workflows/BuildDylib.yaml
```diff
@@ -2,14 +2,11 @@ name: Build Dylib
...
-      - master
-    # tags: '*'
-  release:
-    types:
-      - created
-  pull_request:
-    types: [opened, synchronize, reopened]
+    paths:
+    - 'deps/Makefile'
+    - 'deps/libtask.c'
```
devmotion, line 7: My bad, it should be
```suggestion
    - 'deps/task.c'
```

deps/build.jl
```diff
@@ -1,65 +1,83 @@
...
+    if Libdl.dlopen_e(libtask) in (C_NULL, nothing)
+        error("\$(libtask) cannot be opened, Please re-run Pkg.build(\\"Libtask.jl\\"), and restart Julia.")
+    end
+
 end
 
-version_str = get_version_str() |> strip |> (x) -> lstrip(x, ['v'])
-include_build_script(version_str, true)
+"""
+write(joinpath(@__DIR__, "deps.jl"), deps_source)
```
devmotion, line 83: I still think if possible we should not write `deps.jl` explicitly. For instance, if you compare this with the implementation of `write_deps_file` in https://github.com/JuliaPackaging/BinaryProvider.jl/blob/81f03dad25dfd76845f4fbc0eea1a383280f334c/src/Products.jl#L397-L507 which is usually used in builds.jl, you see that in the standard version there are many more checks and a much more complicated logic for deriving the correct paths on each OS. IMO if possible we shouldn't try to outsmart this implementation.

If we want a custom `check_deps()` function it might be better to just add a custom check in src/Libtask.jl after including `deps/deps.jl` and not running `check_deps()`.

devmotion:
So I think the following approach could work:
- Use the generated build.jl and just add `prefix` to the LibraryProduct definitions
- Define the `const libtask` as described above in src/Libtask.jl
- in the `__init__()` method in src/Libtask.jl just check if `libtask` can be loaded in the same way as it's currently done in deps/deps.jl, i.e., copy the check that you added to the custom build script.

In that way we don't mess with the generation of the deps.jl file (in particular not with paths) but still just check and use the version that we want to use.

src/Libtask.jl
```diff
@@ -5,7 +5,24 @@ export CTask, consume, produce, TArray, get, tzeros, tfill, TRef
...
+        const libtask = libtask_v1_3
+    end
+
+    if !isfile(libtask)
+        error("$(libtask) does not exist, Please re-run Pkg.build(\"Libtask.jl\"), and restart Julia.")
+    end
+
+    if Libdl.dlopen_e(libtask) in (C_NULL, nothing)
+        error("$(libtask) cannot be opened, Please re-run Pkg.build(\"Libtask.jl\"), and restart Julia.")
+    end
```
devmotion, line 25: I suggest you move the checks to an `__init__()` method so that they are only executed once after the module has been loaded. See https://github.com/JuliaPackaging/BinaryProvider.jl/blob/master/test/LibFoo.jl/src/LibFoo.jl for an example setup.

BTW with the additional definitions of `libtask` inow it might might be easier to just check if the file does not exist, as shown in the example linked above, and remove the second branch, i.e., just execute everything if it didn't error.

KDr2:
I am sorry that this way doesn't work...

`build.jl` also checks the libs using `dlopen` in the function `write_deps_file`, if it fails on macOS, it won't write `deps.jl`. This is the error: https://github.com/TuringLang/Libtask.jl/pull/58/checks?check_run_id=554895456#step:4:35

So we have to modify `build.jl` to avoid calling `write_deps_file`.

devmotion:
Yes due to line https://github.com/JuliaPackaging/BinaryProvider.jl/blob/55aeaea2652c1b43ba882c2f833196e99f10e2db/src/Products.jl#L418

IMO then the second best solution would be to filter the correct product before calling `write_deps_file`. I really think copying or modifying the implementation of `write_deps_file` should be avoided at any means if possible.

KDr2:
> IMO then the second best solution would be to filter the correct product before calling `write_deps_file`. I really think copying or modifying the implementation of `write_deps_file` should be avoided at any means if possible.

OK, they don't make a big difference to me, I will filter the products list before call `write_deps_file` :)

src/taskcopy.jl
```diff
@@ -1,15 +1,5 @@
...
-@static if VERSION < v"1.0.9999" # (-, v1.1)
-    const libtask = libtask_v1_0
-elseif VERSION < v"1.1.9999" # [v1.1, v1.2)
-    const libtask = libtask_v1_1
-elseif VERSION < v"1.2.9999" # [v1.2, v1.3)
-    const libtask = libtask_v1_2
-else # [v1.3, +)
-  const libtask = libtask_v1_3
-end
-
```
devmotion, line 12: I guess you still want to keep this, but possibly in Libtask.jl?

KDr2: Yeah, had gotten them back along with the products filter.

deps/build.jl
```diff
@@ -1,65 +1,58 @@
...
+]
 
-# modify build-tmp.jl to only check correct version libs
-function install_products_filter(build_file)
-    prod_filter = raw"""products_tmp = filter(products) do prod
+products_tmp = filter(products) do prod
     endswith(prod.libnames[1], "$(VERSION.major)_$(VERSION.minor)")
 end
 length(products_tmp) == 0 && (products_tmp = [products[end]])
 products = products_tmp
```
devmotion, line 25: Looks good, I guess we don't have to care about version jumps? In that case, something like
```suggestion
lastminor = findlast(products) do p
    x = match(r"v(\d)_(\d)$", string(p.variable_name))
    x !== nothing && VERSION.major >= parse(Int, x.captures[1]) && VERSION.minor >= parse(Int, x.captures[2])
end
products = products[[lastminor]]
```
would provide a the last compatible version.

src/Libtask.jl
```diff
@@ -3,12 +3,25 @@ module Libtask
...
+end
+
+@static if VERSION < v"1.0.9999" # (-, v1.1)
+    const libtask = libtask_v1_0
+elseif VERSION < v"1.1.9999" # [v1.1, v1.2)
+    const libtask = libtask_v1_1
+elseif VERSION < v"1.2.9999" # [v1.2, v1.3)
+    const libtask = libtask_v1_2
+else # [v1.3, +)
+  const libtask = libtask_v1_3
```
devmotion, line 23:
```suggestion
  const libtask = libtask_v1_3
```

KDr2:
Thank you @devmotion.

PR #61, opened 2020-04-16
COW DATA based on code generation

KDr2:


Project.toml
```diff
@@ -7,7 +7,9 @@ version = "0.3.4"
 
 [deps]
 BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
+IRTools = "7869d1d1-7146-5819-86e3-90919afe41df"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
```
devmotion, line 12: Do we really need MacroTools? It's a quite heavy dependency and often not very performant, which lead to the creation of lighter packages such as https://github.com/invenia/ExprTools.jl or dropping it as a dependency in MuladdMacro (https://github.com/SciML/MuladdMacro.jl/pull/15).

src/cow.jl
```diff
@@ -0,0 +1,141 @@
+using IRTools
+using MacroTools
+
+using IRTools: BasicBlock, stmt, var
+
+const MUTATING_OPS = (:push!, :pop!, :setindex!)
+
+function copy_object(obj)
+    ct = _current_task()
+    ct.storage == nothing && (ct.storage = IdDict())
```
devmotion, line 10: Often it's better to compare `nothing` (and singletons in general) with `===`:
```suggestion
    ct.storage === nothing && (ct.storage = IdDict())
```

src/cow.jl
```diff
@@ -0,0 +1,141 @@
...
+    end
+
+    new_obj = deepcopy(obj)
+    ct.storage[:_cow_asset][new_obj] = true
+    ct.storage[:_cow_copylog][objectid(obj)] = new_obj
+    return new_obj
+end
+
+maybe_copy(x) = x
+maybe_copy(x::AbstractArray{T}) where {T<:Number} = copy_object(x)
```
devmotion, line 32: I guess we don't need the static type parameter here?
```suggestion
maybe_copy(x::AbstractArray{<:Number}) = copy_object(x)
```

src/cow.jl
```diff
@@ -0,0 +1,141 @@
...
+    end
+
+    return ir
+end
+
+insert_copy(ir) = ir |> insert_copy_stage_1 |> insert_copy_stage_2
+
+IRTools.@dynamo function cow(a...)
+    ir = IRTools.IR(a...)
+    ir == nothing && return
```
devmotion, line 130: Same here:
```suggestion
    ir === nothing && return
```

src/cow.jl
```diff
@@ -0,0 +1,141 @@
...
+    ir_new = MacroTools.prewalk(ir) do x
+        if (x isa IRTools.Variable) && haskey(replacements, x)
+            return replacements[x]
+        end
+        return x
+    end
+
+    for (mv, newk) in replacements
+        # ir_new[newk] = :(Libtask.maybe_copy($(ov)))
+        ir_new[newk] = IRTools.xcall(GlobalRef(Libtask, :maybe_copy), mv)
```
phipsgabler, line 80: IIRC, this method should work as well:

```suggestion
        ir_new[newk] = IRTools.xcall(Libtask, :maybe_copy, mv)
```

test/cow_data.jl
```diff
@@ -0,0 +1,55 @@
...
+
+        ir = IRTools.@code_ir f_ct()
+        @show ir
+        ir = Libtask.insert_copy(ir)
+        print("--------\n")
+        @show ir
+    end
+
+
+    @testset "Test on Array" begin
```
devmotion, line 26: Maybe one could just use the example in the README, that seems a bit simpler?
```julia
function f()
  t = Array{Int}(undef, 1)
  t[1] = 0
  while true
    produce(t[1])
    t[1] = 1 + t[1]
  end
end

t = CTask(f)

@test consume(t) == 0
@test consume(t) == 1

a = copy(t)
@test consume(a) == 2
@test consume(a) == 3

@test consume(t) == 2
@test consume(t) == 3
```

BTW I guess the following works as well?
```julia
function g()
  s = Array{Int}(undef, 1)
  t = s
  t[1] = 0
  while true
    produce(t[1])
    t[1] = 1 + t[1]
  end
end

t = CTask(g)

@test consume(t) == 0
@test consume(t) == 1

a = copy(t)
@test consume(a) == 2
@test consume(a) == 3

@test consume(t) == 2
@test consume(t) == 3
```

src/cow.jl
```diff
@@ -0,0 +1,141 @@
...
+        ir[v] = st_new
+    end
+end
+
+function insert_copy_stage_2(ir::IRTools.IR)
+    mutate_instructions = Dict{IRTools.Variable, IRTools.Variable}()
+
+    for (v, st) in ir
+        isa(v, IRTools.Variable) || continue
+        isa(st, IRTools.Statement) || continue
```
phipsgabler, line 107: I think this is unnecessary.  `iterate` of `IR` only gives you the pairs of variables and statements, no branches.

src/cow.jl
```diff
@@ -0,0 +1,141 @@
...
+        mut || continue
+        if mpos == 1 && isa(st.expr.args[2], IRTools.Variable) # push!(...)
+            mutate_instructions[v] = st.expr.args[2]
+        elseif mpos == 2 && isa(st.expr.args[3], IRTools.Variable)  # self(push!, ...)
+            mutate_instructions[v] = st.expr.args[3]
+        end
+    end
+
+    for (mres, mv) in mutate_instructions
+        rk = insert!(ir, mres, IRTools.xcall(GlobalRef(Libtask, :maybe_copy), mv))
```
phipsgabler, line 118: (See above)

src/cow.jl
```diff
@@ -0,0 +1,141 @@
...
+    return ir_new
+end
+
+function update_var(ir, start, from, to)
+    started = false
+    for (v, st) in ir
+        if v == start
+            started = true
+        end
+        started || continue
```
phipsgabler, line 92:
```suggestion
        v != start && continue
```

src/cow.jl
```diff
@@ -0,0 +1,141 @@
...
+    for (v, st) in ir
+        isa(v, IRTools.Variable) || continue
+        isa(st, IRTools.Statement) || continue
+        mut, mpos = mutating(st.expr)
+        mut || continue
+        if mpos == 1 && isa(st.expr.args[2], IRTools.Variable) # push!(...)
+            mutate_instructions[v] = st.expr.args[2]
+        elseif mpos == 2 && isa(st.expr.args[3], IRTools.Variable)  # self(push!, ...)
+            mutate_instructions[v] = st.expr.args[3]
+        end
```
phipsgabler, line 114:
```suggestion
        if isa(st.expr.args[mpos + 1], IRTools.Variable)
            mutate_instructions[v] = st.expr.args[mpos + 1]
        end
```

src/cow.jl
```diff
@@ -0,0 +1,141 @@
+using IRTools
+using MacroTools
+
+using IRTools: BasicBlock, stmt, var
+
+const MUTATING_OPS = (:push!, :pop!, :setindex!)
```
phipsgabler, line 6: If generalization beyond these is planned, this should contain pairs of the ops and the mutated argument positions, I think

src/cow.jl
```diff
@@ -0,0 +1,141 @@
...
+    return (false, 0)
+end
+
+function insert_copy_stage_1(ir::IRTools.IR)
+    mutate_vars = Dict{IRTools.Variable, Bool}()
+    replacements = Dict{IRTools.Variable, IRTools.Variable}()
+
+    for (v, st) in ir
+        isa(v, IRTools.Variable) || continue
+        isa(st, IRTools.Statement) || continue
```
phipsgabler, line 55: (See below)

Project.toml
phipsgabler: There are MacroTools functions overloaded by IRTools, so it's an indirect depencency anyway.  But I don't think whether its necessary to `using` it when you already use IRTools...

src/cow.jl
phipsgabler: (Or above, as review comments are apperently orderd by time, not order in file?)

Project.toml
phipsgabler: There are MacroTools functions overloaded by IRTools, so it's an indirect depencency anyway. But I don't think whether its necessary to using it when you already use IRTools...

phipsgabler: There are currently two usages of `prewalk`: one is over IR anyway, and the other, in line 94, can surely be replaced by something that neither needs MacroTools, nor ExprTools (expressions in statements are guaranteed to be non-nested anyway).

phipsgabler: It seems my brain likes to use the word _anyway_ a lot.

phipsgabler:
Also, in general, since you only replace single expressions and leave the rest of the code intact, it might be possible to write the whole transformation as one loop over a `Pipe`, which takes care of the subsequent renaming of the variables below.  (You then just have to tell it manually which one you want to replace the old one.)

phipsgabler:
Another architectural consideration -- which might not work in case I misunderstood the idea.

What about replacing the called functions, instead of adding `maybe_copy` into the IR?  Like, instead of transforming

```
%j = push!(%i, args...)
```

into

```
%k = maybe_copy(%i)
%j = push!(%k, args...)
```

where `push!` is one of the "chosen candidates", have _all functions_ be replaced to

```
%j = make_cow(push!, %i, args...)
```

where

```
make_cow(::typeof(push!), x, args...) = push!(maybe_copy(x), args...)
```

for mutating functions, and otherwise

```
make_cow(f, args...) = cow(f, args...)
```

That's actually like own `overdub` implementation, then (just simpler to read).  But it makes extension much nicer.

src/cow.jl
KDr2: These two versions have different meaning:
- `v != start && continue` only picks one statement from the IR (v)
- The multi-line version picks multiple statements from the IR (v and ones after that)

KDr2:
Hi all, I'd update the PR according to your review suggestions, though there are still many problems to fix.

The biggest one I found is that the current approach does not work well in some situations.

Currently, I just do simply replacement on in-place operations, it is transforming 

```
%j = push!(%i, args...)
```

into

```
%k = maybe_copy(%i)
%j = push!(%k, args...)
```

or as @phipsgabler suggested, transforming it into `%j = make_cow(push!, %i, args...)` may have the same effect.

But see this scenario:

```
1:
%1 = make array
%2 = other ops like printing some logs

2:
%3 = getindex(%1, i)
%4 = produce(%3) or give %3 to someone else to use, # this ops may cuase the task to be scheduled off
%5 = push!(%1, ...)
br 2 # for simplicity this is a dead loop

3:
return
```

Will be transformed to

```
1:
%1 = make array
%2 = other ops like printing some logs

2:
%3 = getindex(%1, i)
%4 = produce(%3) or give %3 to someone else to use, # this ops may cuase the task to be scheduled off
%100 = maybe_copy(%1)  # <-updated
%5 = push!(%100, ...)  # <-updated
br 2 # for simplicity this is a dead loop

3:
return
```

The result is not right:
1. enter block 2, produce %1[i]
2. copy %1 to %100 if the task had been copied at %4
3. push!
4. enter block 2 for the second time, produce %1[i], which is a mistake: we should produce %100[i] but not %1[i]


I am not sure if the approach below can fix this problem:

1. find blocks that mutate variables, in this example, **block 2 updates %1**
2. find the successor block of the blocks found in step 1, here, block 2 is a successor of block 2
3. for a given variable, we insert a `maybe_copy` function call before every statement the var is accessed (both read and write)

In this way, the result IR of the example will like

```
1:
%1 = make array
%2 = other ops like printing some logs

2:
%100 = maybe_copy(%1)  # <-updated
%3 = getindex(%100, i) # <-updated
%4 = produce(%3) or give %3 to someone else to use, # this ops may cuase the task to be scheduled off
%101 = maybe_copy(%1)  # <-updated
%5 = push!(%101, ...)  # <-updated
br 2 # for simplicity this is a dead loop

3:
return
```

Any suggestion @phipsgabler?

KDr2:
BTW, calling `maybe_copy(%1)`  in a certain task will get the same object.

src/cow.jl
phipsgabler: Oh yeah, right, that was a stupid suggestion.  It just stroke me an unaesthetic.  Too bad that there is no `dropwhile` in the standard library.

phipsgabler:
Unfortunately I have never worked with Libtask, so I'm still unclear about the semantics.  But maybe it would help if you transformed the whole IR to a structure where `produce` is always at a certain position within a block, like a branch -- either at the beginning or the end.  And make all implicit branches explicit (with `IRTools.explicitbranch!`):

```
1:
%1 = make array
%2 = other ops like printing some logs
br 2

2:
%3 = getindex(%1, i)
%4 = produce(%3) or give %3 to someone else to use, # this ops may cuase the task to be scheduled off
br 3

3:
%100 = maybe_copy(%1)  # <-updated
%5 = push!(%100, ...)  # <-updated
br 2 # for simplicity this is a dead loop
```

test/cow_data.jl
KDr2: Done.

KDr2:
I think it works (on Array{Number} and Dict) now. I also added a `cow` parameter to `CTask` whose default value is `true`, should we set this parameter `false`?

src/cow.jl
```diff
@@ -0,0 +1,219 @@
...
+const MUTATING_OPS = Dict{Symbol, Int}(
+    :setindex! => 1,
+    :push! => 1, :pushfirst! => 1,
+    :pop! => 1, :popfirst! => 1,
+    :append! => 1,
+    :delete! => 1, :deleteat! => 1,
+    :setdiff! => 1,
+)
+
+function mutating(e)
```
yebai, line 47: Pls, consider adding some docs on the function, e.g. functionality, input and output.

src/cow.jl
```diff
@@ -0,0 +1,219 @@
...
+        ct.storage[:_cow_asset] = IdDict{Any, Bool}()
+    end
+    if !haskey(ct.storage, :_cow_copylog)
+        ct.storage[:_cow_copylog] = Dict{UInt64, Any}()
+    end
+
+    haskey(ct.storage[:_cow_asset], obj) && (return obj) # own this obj
+    if haskey(ct.storage[:_cow_copylog], objectid(obj)) # had copied this obj
+        ret = ct.storage[:_cow_copylog][objectid(obj)]
+        ct.storage[:_cow_asset][ret] = true
```
yebai, line 17: I don't fully get the logic here: Task `storage` dict gets shallowly copied during task copying (see [here](https://github.com/TuringLang/Libtask.jl/blob/master/src/taskcopy.jl#L65)), so all cloned task will have the same `storage`. This means all cloned task will share `_cow_asset` and `_cow_copylog`. Right?

src/cow.jl
```diff
@@ -0,0 +1,219 @@
...
+
+    # IRTools.Statement(:((IRTools.Self())(Main.push!, %2, 1)), Any, 2)
+    if (e.args[1] === IRTools.self) && isa(e.args[2], GlobalRef) &&
+        haskey(MUTATING_OPS, e.args[2].name)
+        return (true, MUTATING_OPS[e.args[2].name] + 1)
+    end
+    return (false, 0)
+end
+
+function all_successors(b::IRTools.Block, accu::Vector{Int64})
```
yebai, line 62:
```suggestion
function _successors(b::IRTools.Block, accu::Vector{Int64})
```

src/cow.jl
```diff
@@ -0,0 +1,219 @@
...
+    succs = IRTools.successors(b)
+    ids = map(x -> x.id, succs)
+    issubset(ids, accu) && return
+    for blk in succs
+        push!(accu, blk.id)
+        all_successors(blk, accu)
+    end
+end
+
+function all_successors(b::IRTools.Block)
```
yebai, line 72:
```suggestion
function _successors(b::IRTools.Block)
```

src/cow.jl
```diff
@@ -0,0 +1,219 @@
...
+    ret = Int64[]
+    all_successors(b, ret)
+    return ret
+end
+
+function find_mutating_blocks(ir::IRTools.IR)
```
yebai, line 78: This function finds all IR blocks that contain a mutating operation, right? Also, pls add some docs here.

src/cow.jl
KDr2: https://github.com/TuringLang/Libtask.jl/blob/cow/src/taskcopy.jl#L66

`_cow_asset` and `_cow_copying` is reset while copying, so the new cloned task owns nothing in its asset.

src/cow.jl
KDr2: Yes, docs added.

src/cow.jl
KDr2: done.

src/cow.jl
KDr2: done.

src/cow.jl
```diff
@@ -71,117 +93,125 @@ maybe_copy(x::AbstractDict) = begin
...
+- `print(a)` -> (false, [0])
 
 """
 function mutating(e)
-    isa(e, Expr) || return (false, 0)
-    e.head == :call || return (false, 0)
-    if isa(e.args[1], GlobalRef) && haskey(MUTATING_OPS, e.args[1].name)
-        return (true, MUTATING_OPS[e.args[1].name])
+    isa(e, Expr) || return (false, [0])
+    e.head == :call || return (false, [0])
```
phipsgabler, line 120: You could use `Meta.isexpr(expr, :call)` and an `if` block around the whole inner code, instead.

src/cow.jl
```diff
@@ -71,117 +93,125 @@ maybe_copy(x::AbstractDict) = begin
...
-    issubset(ids, accu) && return
-    for blk in succs
-        push!(accu, blk.id)
-        _successors(blk, accu)
+    if isa(args[1], GlobalRef)
+        if haskey(MUTATING_OPS, args[1])
+            return (true, MUTATING_OPS[args[1]] .+ offset)
+        elseif haskey(MUTATING_OPS, args[1].name)
+            return (true, MUTATING_OPS[args[1].name] .+ offset)
+        end
```
phipsgabler, line 134:
```suggestion
    fn = isa(args[1], GlobalRef) ? args[1].name : args[1]
    if haskey(MUTATING_OPS, fn)
        return (true, MUTATING_OPS[fn] .+ offset)
    end
```

Or even have a separate `remove_global` function, if this is needed more than once.

src/cow.jl
```diff
@@ -304,10 +341,18 @@ end
 
 # debug
 function cmp_cow_ir(func, args...)
```
phipsgabler, line 343: Maybe something like [this macro](https://github.com/TuringLang/IRTracker.jl/blob/c1987c753a86557e79e38a74abdb342d83320f63/src/tracker.jl#L266) would be nice for debugging.  Like Zygote, that has its `@code_adjoint` as well.

src/cow.jl
KDr2: Well, the two versions are not the same as what they are doing.
This piece of bad-smelling code is here because I put both symbols and `GlobalRef`s in `MUTATING_OPS`:

```
MUTATING_OPS = Dict(
:push! => [...],
GlobalRef(Mod1, :fund) => [...]
)
```
I wanted to make all the keys `GlobalRef`, but I find the function `Base.push!` is sometimes referred to as `AnotherModuleThanBase.push!`, so I can't tell if it is `Base.push!`, is there an elegant way to deal with this?

Thanks @phipsgabler

src/cow.jl
devmotion: Or alternatively just
```suggestion
    Meta.isexpr(expr, :call) || return (false, [0])
```
if you don't want to wrap everything in an `if` block.

yebai:
Let's open another PR for approach 2 as discussed.

PR #63, opened 2020-04-27
COW: Another approach.

KDr2:


src/cow.jl
```diff
@@ -0,0 +1,201 @@
...
+end
+
+"""
+    @cow_dispatch(func, arg_positions)
+
+Redispatch methods of a functon based on the types of its arguments.
+`@cow_dispatch(Base.push!, [1])` will generate:
+
+``` julia
+COW_DISPATCH_INFO[:(Base.push!)] = [1]
```
yebai, line 40: I find the logic a bit hard to follow here. Perhaps simplify by 

- Removing the macro ` @cow_dispatch`, and directly initialize the content of `COW_DISPATCH_INFO`?
- Introduce a `DeepCopy` context, then directly defining the following method without the `Base.push!(Val{:COW}, ...)` method with the `Vla{:COW}` nuisance argument:

```julia

for F in (:getindex, :iterate, :eltype, :length, :size,
          :firstindex, :lastindex, :ndims, :axes)
       @eval  (dc::DeepCopy)(::typeof(Base.$F), args...) = function Base.$F(::Type{Array{T, N}},
                           array, args...) where {T, N}
        obj = obj_for_reading(array)
        return $F(obj, args...)
    end
end
```

Context example: https://mikeinnes.github.io/IRTools.jl/latest/dynamo/#Contexts-1

src/cow.jl
```diff
@@ -0,0 +1,201 @@
...
+    quote
+        Libtask.COW_DISPATCH_INFO[$(expr_func)] = $(arg_positions)
+        Libtask.cow(::typeof($(func)), args...) =
+            $(func)(Val(:COW), $(type_args...), args...)
+        $(func)(::Val{:COW}, $(types...), args...) = $(func)(args...)
+    end |> esc
+end
+
+
+"""
```
yebai, line 65: I am slightly confused. Isn't `non_cow_func` similar with `COW_DISPATCH_INFO`? Can we merge these two?

src/cow.jl
```diff
@@ -0,0 +1,201 @@
...
+        (0, obj)
+    end
+    return d
+end
+obj_for_reading(obj) = _obj_for_reading(obj)
+Base.get(obj) = obj_for_reading(obj)
+
+function _obj_for_writing(obj)
+    oid = objectid(obj)
+    n, d = try
```
yebai, line 126: this `try` block can be simplified, see `? get!`

src/ctask.jl
```diff
@@ -33,11 +33,11 @@ proper way is refreshing the `current_task` (the variable `t`) in
 `start_task` after the call to `jl_apply` returns.
 
 """
-function task_wrapper(func)
+function task_wrapper(func; _cow=true)
```
yebai, line 36: maybe keep the keyword argument consistent with `CTask`?

```suggestion
function task_wrapper(func; cow=true)
```

src/cow.jl
KDr2: It's a `get` not `get!` as it only tries to read the local storage. I use this because I didn't want to rely on `task.storage` which is a hidden implementation detail. now I changed it to:
```
 get(Base.get_task_tls(task), (0, obj))
```

src/cow.jl
```diff
@@ -0,0 +1,201 @@
...
+        ir[x] = Expr(:call, to, st.expr.args...)
+    end
+    return ir
+end
+
+IRTools.@dynamo function cow(a...)
+    ir = IRTools.IR(a...)
+    ir === nothing && return
+    # IRTools.recurse!(ir)
+    recurse_no_builtin!(ir)
```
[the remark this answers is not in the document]
KDr2: done.

src/ctask.jl
KDr2: If we use `cow` in the arguments, it will shadow the function `cow`. 

Now I change the argument to `cow` and the function to `_cow`.

src/cow.jl
KDr2: I think it is the same to use a Context (closure) or a simple function here. The only confusion is why there is a `Val{:COW}` in the argument list. I will explain this later.

`@cow_dispatch` does 3 things (let's use `@cow_dispatch(Base.push!, [1])` as an example):
-  `COW_DISPATCH_INFO[:(Base.push!)] = [1]`: Putting the function name in `COW_DISPATCH_INFO`, this info let us know we should skip `Base.push!` in the function `_recurse!`, so any function calls within `Base.push!` will not apply the COW mechanism.
-  `_cow(::typeof(Base.push!), args...) = Base.push!(Val(:COW), typeof(args[1]), args...)`: Add a method to `_cow`, so we can capture every call to `Base.push!` and redirect the call to `Base.push!(Val(:COW), typeof(args[1]), args...)`.
I use `Base.push!(Val(:COW), typeof(args[1]), args...)` but not `Base.push!(typeof(args[1]), args...)` because `Base.func(typeof(args[1]), args...)` is an established function in the stdlib, like ` function getindex(::Type{T}, vals...) where T` in `Base`, using a `Val{:COW}` as the first argument can avoid this kind of method dispatch conflict (or ambiguity). 
- ` Base.push!(::Val{:COW}, ::Type, args...) = Base.push!(args...)`: define a general fallback function, so if we call `Base.push!` on a non-COW type in a COW function (ctask), like `Base.push!(q::Queue, 1)`, the call becomes `Base.push!(Val(:COW), Queue, q, 1)` then fallback to `Base.push!(q,1)`


Directly initializing the content of COW_DISPATCH_INFO then defining `_cow(::typeof(Base.$F), args...)` for each function is corresponding to step 1 and step 2, but leaves step 3 behind.

I think the macro `@cow_dispatch` really simplify the processes we implement COW on a data type:
1. make sure all the reading/writing functions are `@cow_dispatch`ed
2. override these methods for your data type

src/cow.jl
KDr2: - It does step 2 of `@cow_dispatch`, step 1 and 3 are not necessary for this use (only make a function non-cow, but not hooking it to modify its behavior)
- it can make named functions, anonymous functions, closures, etc non-cow, `@cow_dispatch` and `COW_DISPATCH_INFO` only works for named functions.

KDr2:
rebased.

.github/workflows/Testing.yaml
```diff
@@ -11,25 +11,48 @@ on:
...
           - '1.4'
-          - 'nightly'
         os:
           - ubuntu-latest
           - windows-latest
           - macOS-latest
         arch:
           - x64
           - x86
+        nightly: [false]
```
yebai, line 31: the keyword `nightly` is ambiguous, maybe replace with:

```suggestion
        is_nightly: [false]
```

KDr2: done.

src/cow.jl
yebai: > Directly initializing the content of COW_DISPATCH_INFO then defining _cow(::typeof(Base.$F), args...) for each function is corresponding to step 1 and step 2, but leaves step 3 behind.

I think we might be able to merge step 2 and 3, e.g., in the following way:

```julia

_cow(::typeof(Base.push!), ::Type{AbstractArray}, args...) = new method def # step 2
_cow(::typeof(Base.push!), ::Type{AbstractDict}, args...) = new method def # step 2
_cow(::typeof(Base.push!), ::Type, args...)  = Base.push!(args...) # step 3
```

Right?

yebai:
thanks, @KDr2, I still find the overall logic quite confusing. My impression is that we only need to intercept certain functions for `Array` types, e.g. `getindex`, `setindex!` etc. Then, we call our own implementation of `getindex` and `setindex!` to detect whether task copying has happened, if so, perform corresponding deep copy operations.

KDr2:
> thanks, @KDr2, I still find the overall logic quite confusing. My impression is that we only need to intercept certain functions for `Array` types (similarly for `Dict`), e.g. `getindex`, `setindex!` etc. Then, we call our own implementation of `getindex` and `setindex!` to detect whether task copying has happened, if so, perform corresponding deep copy operations.

Yes, you are right, the PR does exactly what you said.

Function `_cow` is used to intercept the `getindex`, `setindex!` calls, and re-dispatch them. but when intercepting/re-dispatching the calls, we can't know their argument type at compile-time,  so we can't do "if its argument is an Array, then call `array_getindex(...)`", we only can rely on multi-dispatch, call `getindex(obj::Array, ...)`, but this method is defined by Julia. So I defined
-  `getindex(::Type{Array}, obj::Array, ...)`, our implementation of `getindex`, 
- and a fallback for other types: `getindex(::Type, obj, ...)`

With these two methods, I can do `getindex(obj, args..) = getindex(typeof(obj), obj, args...)` to forward the call to our implementation or the default implementation (via the fallback version).

But unfortunately, `getindex(::Type, obj, ...)` is also a Julia builtin method, so again I defined `: `getindex(::Val{:COW}, ::Type, obj, ...)` ...

src/cow.jl
KDr2: This doesn't work for all the functions, take `getindex` for an example:
```
julia> @code_lowered ((a) -> Array{Int}[a])([1])
CodeInfo(
1 ─ %1 = Core.apply_type(Main.Array, Main.Int)
│   %2 = Base.getindex(%1, a)
└──      return %2
)
```
When using `Array{Int}[a]` to construct an `Array{Array{Int}}`, `getindex(::Type{Array{Int}}, ...)` will be called, according to your code, it will be forwarded to `new_method_def` in the first line, and this is not what we want.

Also, in this way, despite the problem above, the method `new_method_def` should have a name different from `Base.getindex`, otherwise, it must be a method redefinition.

KDr2: And, putting types, like `Array`, `Dict`, here also make the code not such generic: If we want to implement COW for a new type, we must add a type here.

Use `_cow(::typeof(Base.push!), args...)  = Base.push!(typeof(args[1]), args...)` does the same thing, but when adding a new type, what's we should do is just adding a new method `Base.push!(::Type{NewType}, ....)`, without knowing what `_cow` is.

KDr2: Considering that `Base.func(::Type{NewType}, ....)` may be a builtin method, we use `Base.func(::Val{:COW}, ::Type{NewType}, ....)` instead.

src/cow.jl
```diff
@@ -0,0 +1,207 @@
...
+        f = () -> begin
+            $(esc(expr))
+        end
+        Libtask.non_cow_call(f)
+    end
+end
+
+@non_cow_func(produce)
+@non_cow_func(consume)
+@non_cow_func non_cow_call(func, args...) = func(args...)
```
yebai, line 122: I think `non_cow_call` can be removed if we merge the macro `non_cow_func` and `non_cow`.

src/cow.jl
KDr2: Done.

src/cow.jl
```diff
@@ -0,0 +1,207 @@
...
+
+"""
+function _recurse!(ir, to=IRTools.self)
+    for (x, st) in ir
+        IRTools.isexpr(st.expr, :call) || continue
+        func = st.expr.args[1]
+        if isa(func, GlobalRef)
+            if func.mod in (Base, Core)
+                fexpr = Expr(:., Symbol(func.mod), QuoteNode(func.name))
+                if !haskey(COW_DISPATCH_INFO, fexpr)
```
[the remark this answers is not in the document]
KDr2: Done.

src/cow.jl
```diff
@@ -0,0 +1,207 @@
...
+IRTools.@dynamo function _cow(a...)
+    ir = IRTools.IR(a...)
+    ir === nothing && return
+    # IRTools.recurse!(ir)
+    _recurse!(ir)
+    return ir
+end
+
+"""
+    @cow_dispatch(func, arg_positions)
```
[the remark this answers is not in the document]
KDr2: Done.

src/cow.jl
```diff
@@ -0,0 +1,207 @@
...
+`@cow_dispatch(Base.push!, [1])` will generate:
+
+``` julia
+COW_DISPATCH_INFO[:(Base.push!)] = [1]
+_cow(::typeof(Base.push!), args...) =
+    Base.push!(Val(:COW), typeof(args[1]), args...)
+Base.push!(::Val{:COW}, ::Type, args...) = Base.push!(args...)
+```
+"""
+macro cow_dispatch(func, arg_positions)
```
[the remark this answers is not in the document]
KDr2: Done.

src/cow.jl
```diff
@@ -0,0 +1,207 @@
...
+
+Make a function skip the COW mechanism.
+
+```julia
+@non_cow_func function f1()
+    # ...
+end
+```
+"""
+macro non_cow_func(func)
```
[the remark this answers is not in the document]
KDr2: Done.

src/cow.jl
```diff
@@ -0,0 +1,207 @@
...
+"""
+    @non_cow
+
+Make an expression skip the COW mechanism.
+
+```julia
+@now_cow data[idx] = value
+```
+"""
+macro non_cow(expr)
```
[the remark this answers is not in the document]
KDr2: Done.

src/cow.jl
```diff
@@ -0,0 +1,207 @@
...
+    quote
+        f = () -> begin
+            $(esc(expr))
+        end
+        Libtask.non_cow_call(f)
+    end
+end
+
+@non_cow_func(produce)
+@non_cow_func(consume)
```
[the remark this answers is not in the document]
KDr2: Yes, we need this, they are not in `COW_DISPATCH_INFO`, so we should use this to skip the interception by `_cow`.

Most methods in Base and Core will/should be skipped the interception, except the ones  we use to read/update COW-DATA-Structures, like push!, pop!, etc. These exceptions are put into CCOW_DISPATCH_INFO.

So functions in COW_DISPATCH_INFO is to say "Though I'm in module Base/Core, please intercept me."

src/cow.jl
KDr2: Not really, but I renamed it to `nevercopy`, so now we have both a macro and a function named `nevercopy`.

We can remove it because here (https://github.com/TuringLang/Libtask.jl/pull/63/files#diff-4f6fdc6f158cda9bca820a96156a34dcR107) we must use it to make the local function `f` skip the _cow interception.

src/cow.jl
```diff
@@ -0,0 +1,203 @@
...
+        newd = deepcopy(d)
+        task_local_storage(oid, (cn, newd))
+    end
+    return newd
+end
+obj_for_writing(obj) = _obj_for_writing(obj)
+
+### COW DISPATCH
+# write
+@maybecopy(Base.setindex!, [1])
```
yebai, line 145: @maybecopy func(a, ::AbstractArray, x, y, z) # @maybecopy(func, [2])

yebai: _maybecopy_func(...)

KDr2: Done.

yebai:
We decided to close this PR after some discussion with @KDr2. We will experiment with an alternative `Alloc` based mechanism in a separate PR.

Issue #64, opened 2020-04-27
Make a CTask struct


devmotion:
As discussed in https://github.com/TuringLang/Turing.jl/pull/1237, it would be nice to implement the iteration interface for `CTask` since then it would be possible to use alternatives such as resumable functions in the particle filter implementations in Turing in a straightforward way (currently `produce` and `consume` statements are hardcoded). However, currently `CTask` is just a function that returns a `Task` for which stack copying has been enabled. Implementing the iteration interface for `Task` would be pretty bad type piracy (BTW `Base.copy(::Task)` in the current code is also type piracy). To avoid this I suggest adding a `CTask` struct that is defined as
```julia
struct CTask
	task::Task
end
```
Then `Base.copy` and `Base.iterate` could be implemented without committing type piracy. The downside would be that one would probably have to forward a bunch of methods defined for `Task` (such `istaskstarted` etc.) to `task`.

Do you think I should give it a try, or are there any other suggestions?

yebai:
yes, I think this is doable, and would be a nice improvement to both Libtask and Turing.

yebai:
closed by #65

PR #65, opened 2020-04-28
(Too) many changes...

devmotion:
I had a closer look at the implementation due to https://github.com/TuringLang/Libtask.jl/issues/64 and ended up with a huge amount of changes, so it might be better to split this PR...

Basically, it includes:
- A `CTask` wrapper for `Task`s for which task copying is enabled
- An implementation of the iteration interface for `CTask` that just consumes
- A cleaner `CTaskException`
- A lot of additional comments in the source code and (hopefully) cleaner syntax
- A cleaner implementation of the test suite, by, e.g., grouping together related tests in test sets and replacing `@assert` with `@test`
- Including the implementation and tests of `TRef`
- Update of the README (while writing the tests I noticed that using the same variable names inside of the function and for the task is a bit unfortunate since it breaks in a local scope, i.e., e.g., inside of `@testset`).


devmotion:
Hmmm seems I introduced a bug in Julia 1.0 :thinking:

src/ctask.jl
```diff
@@ -34,172 +58,154 @@ proper way is refreshing the `current_task` (the variable `t`) in
...
-        isa(ct.storage, Nothing) && (ct.storage = IdDict())
-        ct.storage[:_libtask_state] = :failed
-        wait()
+    f = let func=func
+        () -> begin
+            try
+                ct = _current_task()
+                res = func()
+                ct.result = res
+                ct.state = :done
```
yebai, line 67: this change might be problematic during task switching, although I can't remember the exact details

maybe @KDr2 can clarify?

src/ctask.jl
```diff
@@ -34,172 +58,154 @@ proper way is refreshing the `current_task` (the variable `t`) in
...
+                wait()
+            catch ex
+                ct = _current_task()
+                ct.exception = ex
+                ct.result = ex
+                ct.backtrace = catch_backtrace()
+                ct.state = :failed
```
yebai, line 74: similar here.

src/ctask.jl
devmotion: Hmm yes, I was wondering if it is needed. I removed it (which worked fine on Julia > 1.0 it seems) since in https://github.com/TuringLang/Libtask.jl/blob/2ca2a89a5c8e46d21278ab93cd3caed48e6f7a71/src/ctask.jl#L189-L190 `ct.state` is set to `ct.storage[:_libtask_state]` if it exists. But maybe it's actually needed.

src/ctask.jl
```diff
@@ -34,172 +58,154 @@ proper way is refreshing the `current_task` (the variable `t`) in
...
-        Base.schedule(p)
+    if producer.state === :runnable
+        # Switch to the producer.
+        Base.schedule(producer)
         yield()
 
-        isa(p.storage, IdDict) && haskey(p.storage, :_libtask_state) &&
-            (p.state = p.storage[:_libtask_state])
+        # Throw an exception if the task failed.
+        producer.state === :failed && throw(CTaskException(producer))
```
yebai, line 204: similar here.

src/ctask.jl
devmotion: I'll check what happens if I revert the changes that you commented on.

devmotion: It seems to resolve the issue but I would still be interested to know the details - @KDr2, can you explain why this separate state is needed?

devmotion: And there's still something not quite right it seems.

yebai: Seems not related to `Libtask`, more like a Julia Windows issue on the nightly build

KDr2: > I would still be interested to know the details

Any copy of a task should NOT finish or fail before the original task finishes or fails. So I added a `wait()` at the end of each branch in the function `task_wrapper`, and that makes the tasks and their copies never be done... It's a bad solution but I couldn't find a better way to ensure a task don't end before its origin end.

If we assign `:done` or `:failed` directly to `task.state`, the task ends, and this may happen before its origin ends.

The root reason is how Julia starts and ends a task (pseudo code):
```
jl_task_t *t = get current task; // (a)
t->run_the_julia_code(); // (b) here, we copy the task
finish_and_clean(t); // (c)
```

a. `t` is the original task
b. origin and copies. Let's assume a copy (say, task_ct)  ends first, then the control flow goes to (c)
c. we come back to here because a copy (task_ct) ends, but the original task (t) doesn't ends yet, but `t` is the original task and we do clean job on it...


There are our ways to fix this:
- at the (c) point, we refresh the variable `t` (`t = get_current_task();`) then clean it. this solution is the perfect one but it should modify the code of JuliaLang...
- do some hack, update the t variable (which is on the stack) in a copied task, this is hard and arch specific and not portable...

yebai: thanks, @KDr2. As @devmotion mentioned, removing this hack seems working fine for Julia versions from `v1.1`. It would be interesting to understand why. Also, can you add some notes to this hack in a separate PR for future reference?

KDr2: OK, I will do some investigation and write some docs. But I think it must don't work under Julia >1.1 without the hack neither since I just looked into relevant code in JuliaLang today, maybe the bug is not triggered by current test cases. Anyway, I will write this down and make it clean how it works and how it doesn't work.

devmotion: In any case, thanks for the detailed explanations!

PR #70, opened 2020-05-25
Fix convert

mohamed82008:
This PR fixes some `convert` bugs.

### DIFF
--- a/src/tarray.jl
+++ b/src/tarray.jl
@@ -87,10 +87,10 @@ function Base.pop!(S::TArray)
     pop!(d)
 end
 
-function Base.convert(::Type{TArray{T, N}}, x::Array{T, N}) where {T, N}
-    return convert(TArray, x)
-end
 function Base.convert(::Type{TArray}, x::Array)
+    return convert(TArray{eltype(x),ndims(x)}, x)
+end
+function Base.convert(::Type{TArray{T,N}}, x::Array{T,N}) where {T,N}
     res = TArray{typeof(x[1]),ndims(x)}();
     n   = n_copies()
     task_local_storage(res.ref, (n,x))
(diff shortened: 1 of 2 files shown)


devmotion:
It would be good to add some tests for these methods.

mohamed82008:
Done.

src/tarray.jl
```diff
@@ -87,14 +87,20 @@ function Base.pop!(S::TArray)
     pop!(d)
 end
 
+function Base.convert(::Type{TArray{T, N}}, x::Array{T, N}) where {T, N}
+    return convert(TArray, x)
+end
 function Base.convert(::Type{TArray}, x::Array)
```
devmotion, line 93: Wouldn't it be more natural to define
```suggestion
function Base.convert(::Type{TArray}, x::Array)
    return convert(TArray{eltype(x),ndims(x)}, x)
end
function Base.convert(::Type{TArray{T,N}}, x::Array{T,N}) where {T,N}
```
if `convert(::Type{TArray{T,N}}, ::Array{T,N})` is needed?

mohamed82008: Sure.

src/tarray.jl
```diff
@@ -88,15 +88,21 @@ function Base.pop!(S::TArray)
 end
 
 function Base.convert(::Type{TArray}, x::Array)
-    res = TArray{typeof(x[1]),ndims(x)}();
+    return convert(TArray{eltype(x),ndims(x)}, x)
+end
+function Base.convert(::Type{TArray{T,N}}, x::Array{T,N}) where {T,N}
+    res = TArray{T,N}();
```
devmotion, line 94:
```suggestion
    res = TArray{T,N}()
```

Issue #75, opened 2020-10-09
Use Artifacts instead of BinaryProvider/build.jl


aviks:
It would be useful to use Artifacts and jll packages, instead of creating a build.jl and using Binaryprovider to download the depdendencies. 

Using Artifacts has a couple of advantages: 1) it reduces duplication in download and storage of the native depdendencies, and 2) it works via the package server mechanism, which provides a globally replicated CDN, better caching, and better support for proxies. 

This would mean however that only Julia 1.3 and above are supported, but I noticed that Turing.jl already has 1.3 as minimum bound, so should be fine to limit this pakage to that lower bound.

devmotion:
IMO it would make sense to move away from BinaryProvider and drop support for Julia < 1.2, since it is not supported anymore by Turing (and many other TuringLang packages).

Libtask is a bit special since it requires a specific library for every Julia version (which leads to hickups when e.g. updating Julia, and requires special handling in [build.jl](https://github.com/TuringLang/Libtask.jl/blob/66104ba0cfebdea55a4cc4de77f30bdc118bd45b/deps/build.jl#L14-L23)). Can such a dependency on the Julia version be handled nicely in the Artifacts system? Is the way to go just to build and distribute all libraries for all Julia version and select the correct one at loading time in the same way as in https://github.com/TuringLang/Libtask.jl/blob/66104ba0cfebdea55a4cc4de77f30bdc118bd45b/src/Libtask.jl#L16-L24? Or can this be done more smoothly nowadays?

ViralBShah:
I thought Project.toml can encode the exact versions or range of versions of dependencies for every release.

PR #76, opened 2020-10-09
Use Libtask_jll

devmotion:
This PR fixes https://github.com/TuringLang/Libtask.jl/issues/75 by removing the custom build process with BinaryProvider and instead using the prebuilt libraries in Libtask_jll.

This change requires Julia >= 1.3 but I assume this might be fine since recent versions of Turing only support Julia >= 1.3.


.github/workflows/CompatHelper.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  CompatHelper:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Pkg.add("CompatHelper")
+        run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
+      - name: CompatHelper.main()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}
```
devmotion, line 15: CI tests will only run if an SSH key is added (see https://github.com/JuliaRegistries/CompatHelper.jl#12-set-up-the-ssh-deploy-key-optional).

Project.toml
```diff
@@ -3,16 +3,16 @@ uuid = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 license = "MIT"
 desc = "C shim for task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
-version = "0.4.2"
+version = "0.5.0"
```
devmotion, line 6: This allows us to backport and release bugfixes for Julia < 1.3 in Libtask 0.4.* if needed.

devmotion:
There seems to be some reproducible test errors on Windows32 with Julia 1.5 (and test errors on nightly since apparently internals of Task were changed). Can they be fixed?

Project.toml
```diff
@@ -3,16 +3,16 @@ uuid = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
...
 
 [deps]
-BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
```
aviks, line 9: Is Libdl still required?

devmotion: Just forgot to remove it :slightly_smiling_face:

KDr2:
So where the removed files (c source, makefile, build script, etc) are placed now? I did a simple search and failed to find them...

devmotion:
The intention was to not keep them in the repo here but to move them to Yggdrasil. That would keep everything together in one place and make it easier to update and rebuild the library, without making PRs to two repositories.

BTW there are efforts to link libraries against Julia on Yggdrasil, and it seems there's already official support for Julia 1.4 (but not all OS) and 1.5. So at some point we might be able to get rid of the different libraries and the checks in Libtask.jl, and instead only use a Julia-versioned `libtask` library.

KDr2:
> The intention was to not keep them in the repo here but to move them to Yggdrasil. 

I think it would be better to move the relevant files to Yggdrasil and build a new version of Libtask_jll before merging this PR, what do you think?

> we might be able to get rid of the different libraries and the checks in Libtask.jl, and instead only use a Julia-versioned `libtask` library.

Great, the build process could be dramatically simplified if it is true :)

devmotion:
> I think it would be better to move the relevant files to Yggdrasil and build a new version of Libtask_jll before merging this PR, what do you think?

Sounds reasonable :+1: At the same time, we should maybe build Julia 1.4 and 1.5 specific versions as well? I'm wondering if this could fix the test issues on Windows 32bit with Julia 1.5.

yebai:
@KDr2 let's get this PR merged asap.

yebai (approved):

devmotion:
We have to wait until the libraries are available in the registry.

KDr2:
> We have to wait until the libraries are available in the registry.

How many versions of the JLL should we build in Yggdrasil?  At least 3 I think? As discussed in https://github.com/JuliaPackaging/Yggdrasil/pull/2087, we have to file a PR for each version and wait until all of them are available in the registry。

devmotion:
Yes, exactly. I wanted to test this PR locally with the library for Julia 1.3 first before opening the other PRs. However, the library is not available in the general registry since automerging failed: https://github.com/JuliaRegistries/General/pull/24916

devmotion:
The new library is still not registered but tests passed locally: https://github.com/JuliaRegistries/General/pull/24916

devmotion:
It seems to work :tada: 

The pre-built Libtask_jll is now available for Julia 1.3 (Libtask_jll 0.4.0), Julia 1.4 (Libtask_jll 0.4.1), and Julia 1.5 (Libtask_jll 0.4.2). In the tests the correct version of Libtask_jll is installed for these Julia versions.

Tests for Julia nightly fail since there is no compatible version of Libtask_jll available.

yebai:
> Tests for Julia nightly fail since there is no compatible version of Libtask_jll available.

Maybe consider falling back to most recent Libtask_jll release (e.g. 0.4.2) for the nightly build?

devmotion:
> Maybe consider falling back to most recent Libtask_jll release (e.g. 0.4.2) for the nightly build?

I think Pkg does not allow this. And actually I think it is reasonable since otherwise we basically create "unbounded" Libtask_jll versions and newer Julia versions might just fall back to completely broken libraries. As far as I know, however, it would be possible to add the Libtask_jll repo manually which then circumvents the checks in the registry. At least that was the case in my local tests - Pkg then just prints a warning about incompatible Julia versions but still installs the latest version. We could add this to the CI tests?

yebai:
> We could add this to the CI tests?

Makes sense since we'll know when breaking changes are introduced into the nightly build. 

Also, maybe consider adding a line in the README file, informing potential nightly Julia users how to install manually?

devmotion:
OK, tests run now on Julia nightly as well with the latest version of Libtask_jll but cause a Julia segfault (not too surprising, they were already broken on master since base Julia removed fields of Task etc.). I also added a section to the README.

yebai:
Excellent work, many thanks @devmotion!

devmotion:
@yebai Can you fix https://github.com/TuringLang/Libtask.jl/pull/76#discussion_r502522118 and add a secret for CompatHelper? Or maybe that has been done already?

yebai:
> @yebai Can you fix #76 (comment) and add a secret for CompatHelper? Or maybe that has been done already?

Done - pls give it a try and see whether it works!

Issue #74, opened 2020-09-26
Handling missing


trappmartin:
The following MWE breaks in the current version of Turing.

```julia
using LinearAlgebra
using Turing

# Linear Model
Δt = 0.1
A = [1.0 Δt; 0.0 1.0]
Q = [0.25 * Δt^4 0.5 * Δt^3; 0.5 * Δt^3 Δt^2] + 1e-6 * I

H = [1.0 0.0; 0.0 1.0;]
R = 0.1 * I

@model function hmm(states, ::Type{T} = Float64) where T
    # Initialize arrays for the time-series
    observations = Matrix{T}(undef, 2, length(states))

    # Prior for first step
    states[1] ~ MvNormal([0.0, 0.5], sqrt(0.001))
    observations[:, 1] ~ MvNormal(states[1], R/Δt)

    for t in 2:length(states)
        states[t] ~ MvNormal(A * states[t-1], Δt*Q)
        observations[:, t] ~ MvNormal(states[t], R/Δt)
    end

    return states
end

# Sample a trial from the prior to be used as the ground truth
hmm_prior = hmm(Union{Missing,Vector{Float64}}[missing for _ in 1:100])
states = hmm_prior()

# Condition on the sampled trial 
hmm_conditioned = hmm(states)

# Perform inference (not working)
chain = sample(hmm_conditioned, SMC(), 1000)
```

With the stack trace:

```
ERROR: CTaskException:
Malformed dims
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] TArray{Union{Missing, Array{Float64,1}},1}(::Array{Union{Missing, Array{Float64,1}},1}) at /Users/martin/.julia/packages/Libtask/Zo6uM/src/tarray.jl:36
```

Any idea why this happens?

When replacing `states = hmm_prior()` with `states = Array{Array{Float64}}(hmm_prior())` everything seems fine.

cpfiffer:
This is more of a Libtask issue than anything else, I think. Your fix is a good shortcut for the moment, but Libtask should probably be more robust to arrays with union types (`Union{Missing, Array{Float64}}`, in this case). It's also an issue of `states` not being tightly typed when it's returned by `hmm_prior`, which you could probably fix by adding to the return statement: 

```julia
@model function hmm(states, ::Type{T} = Float64) where T
    # Initialize arrays for the time-series
    observations = Matrix{T}(undef, 2, length(states))

    # Prior for first step
[...]

PR #77, opened 2020-10-11
Fix constructors

devmotion:
This PR fixes the constructors of `TArray`, as explained in https://github.com/TuringLang/Libtask.jl/issues/74#issuecomment-706680984.


cpfiffer:
Man I was wondering why it was dispatching on that method -- good catch!

cpfiffer (approved):

PR #81, opened 2020-11-26
make Libtask compatible with Julia 1.6

KDr2:
Not ready to merge until https://github.com/JuliaPackaging/Yggdrasil/pull/2176 is merged.

### DIFF
--- a/src/ctask.jl
+++ b/src/ctask.jl
@@ -24,7 +24,12 @@ end
 
 function Base.showerror(io::IO, ex::CTaskException)
     println(io, "CTaskException:")
-    showerror(io, getproperty(ex.task, :exception), getproperty(ex.task, :backtrace))
+    bt = @static if VERSION < v"1.6.0-DEV.1145"
+        ct.backtrace
+    else
+        ct.storage[:_libtask_bt]
+    end
+    showerror(io, ex.task.exception, bt)
 end
 
 # Utility function for self-copying mechanism
@@ -37,7 +42,7 @@ function n_copies(t::Task)
 end
 
 function enable_stack_copying(t::Task)
-    state = getproperty(t, :state)
+    state = t.state
     if state !== :runnable && state !== :done
         error("only runnable or finished tasks' stack can be copied.")
     end
@@ -72,11 +77,12 @@ function task_wrapper(func)
                 ct = _current_task()
                 @static if VERSION < v"1.6.0-DEV.1145"
                     ct.exception = ex
+                    ct.backtrace = catch_backtrace()
                 else
                     ct._isexception = true
+                    ct.storage[:_libtask_bt] = catch_backtrace()
                 end
                 ct.result = ex
-                ct.backtrace = catch_backtrace()
                 ct.storage === nothing && (ct.storage = IdDict())
                 ct.storage[:_libtask_state] = :failed
                 wait()
@@ -88,7 +94,7 @@ end
 
 function Base.copy(ctask::CTask)
     task = ctask.task
-    state = getproperty(task, :state)
+    state = task.state
     if state !== :runnable && state !== :done
         error("only runnable or finished tasks can be copied.")
     end
@@ -139,7 +145,7 @@ function produce(v)
     end
 
     # Internal check to make sure that it is possible to switch to the consumer.
-    @assert getproperty(task, :state) in (:runnable, :queued)
+    @assert task.state in (:runnable, :queued)
 
     @static if VERSION < v"1.1.9999"
         task.state === :queued && yield()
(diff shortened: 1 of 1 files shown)


src/ctask.jl
```diff
@@ -88,7 +94,7 @@ end
 
 function Base.copy(ctask::CTask)
     task = ctask.task
-    state = getproperty(task, :state)
+    state = task.state
     if state !== :runnable && state !== :done
```
devmotion, line 98: Maybe just
```suggestion
    if istaskfailed(task)
```
?

src/ctask.jl
```diff
@@ -139,7 +145,7 @@ function produce(v)
     end
 
     # Internal check to make sure that it is possible to switch to the consumer.
-    @assert getproperty(task, :state) in (:runnable, :queued)
+    @assert task.state in (:runnable, :queued)
 
     @static if VERSION < v"1.1.9999"
```
devmotion, line 150: Remove this branch?
```suggestion
```

src/ctask.jl
```diff
@@ -139,7 +145,7 @@ function produce(v)
     end
 
     # Internal check to make sure that it is possible to switch to the consumer.
-    @assert getproperty(task, :state) in (:runnable, :queued)
+    @assert task.state in (:runnable, :queued)
```
devmotion, line 148:
```suggestion
    @assert !istaskdone(state) && !istaskfailed(task)
```

src/ctask.jl
```diff
@@ -37,7 +42,7 @@ function n_copies(t::Task)
 end
 
 function enable_stack_copying(t::Task)
-    state = getproperty(t, :state)
+    state = t.state
     if state !== :runnable && state !== :done
```
devmotion, line 46: Maybe just
```suggestion
    if istaskfailed(t)
```

src/ctask.jl
devmotion: BTW `_istaskfailed` can be removed since we only support Julia >= 1.3.

src/ctask.jl
```diff
@@ -203,7 +209,7 @@ function consume(ctask::CTask, values...)
         push!(producer.storage[:consumers].waitq, ct)
     end
 
-    if getproperty(producer, :state) === :runnable
+    if producer.state === :runnable
```
devmotion, line 212: Maybe also here use something that is less dependent on the internal state?
```suggestion
    if !istaskdone(producer) && !istaskfailed(producer)
```

src/ctask.jl
KDr2: Done.

src/ctask.jl
KDr2: There's another potential state (`:queued`) here, so I think using `=== :runnable` is more precise here.

src/ctask.jl
KDr2: Done.

src/ctask.jl
devmotion: I think it was removed already before Julia 1.3, so it should not matter anymore.

devmotion: I couldn't find it in https://github.com/JuliaLang/julia/blob/v1.3.0-alpha/base/task.jl.

devmotion: Another indication is this check in ctask.jl: https://github.com/TuringLang/Libtask.jl/blob/a664d3dc5ee8e56ab85dc9368e94f68c8b150083/src/ctask.jl#L144-L146 Apparently it was a thing only in Julia < 1.2.

src/ctask.jl
KDr2: Done.

src/ctask.jl
KDr2: You are right, updated.

devmotion (commented):
Looks good to me but I guess we have to wait until Julia 1.6 is released (or at least one can build binaries for Julia 1.6 on Yggdrasil).

On the other hand, it does not break anything on Julia < 1.6 and only improves the code, so we could also merge it now. It is just not guaranteed that these fixes are sufficient for Julia 1.6.

KDr2:
> On the other hand, it does not break anything on Julia < 1.6 and only improves the code, so we could also merge it now.

Agree.

>  It is just not guaranteed that these fixes are sufficient for Julia 1.6.

I ran the unit tests locally (on linux) with the updated C source, all of them passed.

So it seems that this PR can be merged now.

devmotion:
Good I trust you :slightly_smiling_face: And if we notice any problems in the future, we can always make a bugfix release.

I guess it would be useful to make a release right away if people want to use it with Julia master (even though I don't think it's very likely since they would have to build everything manually)?

devmotion (approved):

yebai (approved):
Great work, thanks @KDr2!

PR #82, opened 2020-12-29
Complete Array API for `TArray`

KDr2:


src/tarray.jl
```diff
@@ -195,3 +101,193 @@ function tfill(val::Real, dim)
...
+
+Base.print_array(io::IO, x::TArray) = Base.print_array(io, get(x))
+
+#
+# Iterator Interface
+#
+
+IteratorSize(::Type{TArray{T, N}}) where {T, N} = HasShape{N}()
+IteratorEltype(::Type{TArray}) = HasEltype()
+
```
devmotion, line 160: I would assume that this is defined automatically since `TArray{T,N} <: AbstractArray{T,N}` - is this not the case?

src/tarray.jl
```diff
@@ -195,3 +101,193 @@ function tfill(val::Real, dim)
...
+#
+# Forward many methods to the underlying array
+#
+for F in (:iterate, :eltype, :length, :size,
+          :firstindex, :lastindex, :ndims, :axes,
+          :collect)
+    @eval Base.$F(a::TArray, args...) = $F(get(a), args...)
+end
```
devmotion, line 168: Same here, I would assume that `ndims` and `eltype` are defined automatically. I think one has only to define `size` and `similar(::TArray, ::Type{T}, dims)`, and probably should define `IndexStyle` and `axes` to exploit the structure of the underlying array. Additionally, maybe `length` is useful, but it is not needed since the default is `length(x) = prod(size(x))`. See https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array for more explanations.

src/tarray.jl
```diff
@@ -195,3 +101,193 @@ function tfill(val::Real, dim)
...
+
+#
+# Similarity implementation
+#
+
+Base.similar(x::TArray) = tzeros(eltype(x), size(x))
+Base.similar(x::TArray, ::Type{T}) where {T} = tzeros(T, size(x))
+Base.similar(x::TArray, dims::Dims) = tzeros(eltype(x), dims)
```
devmotion, line 176: As mentioned above, one should just define
```suggestion
Base.similar(x::TArray, ::Type{T}, dims::Dims) = TArray(similar(get(x), T, dims))
```

BTW in my opinion `get` is really a misuse of `Base.get` which is intended to retrieve a value from a collection or instead a default value if the key does not exist. It would be better to rename it and use an internal function.

src/tarray.jl
```diff
@@ -195,3 +101,193 @@ function tfill(val::Real, dim)
...
+Base.transpose(x::TArray) = transpose(get(x)) |> localize
+Base.adjoint(x::TArray) = adjoint(get(x)) |> localize
+Base.repeat(x::TArray; kw...) = repeat(get(x); kw...) |> localize
+Base.hcat(x::TArray, rest...) = hcat(get(x), _get.(rest)...) |> localize
+Base.hcat(x::AbstractArray, y::TArray, rest...) = hcat(x, get(y), _get.(rest)...) |> localize
+Base.vcat(x::TArray, rest...) = vcat(get(x), _get.(rest)...) |> localize
+Base.vcat(x::AbstractArray, y::TArray, rest...) = vcat(x, get(y), _get.(rest)...) |> localize
+Base.cat(x::TArray, rest...; dims) = cat(get(x), _get.(rest)...; dims = dims) |> localize
+Base.cat(x::AbstractArray, y::TArray, rest...; dims) =
+    cat(x, get(y), _get.(rest)...; dims = dims) |> localize
```
devmotion, line 242: I am a bit worried that some of these dispatches cause method invalidations (it happened before in MCMCChains).

src/tarray.jl
```diff
@@ -195,3 +101,193 @@ function tfill(val::Real, dim)
...
+#
+# Representation
+#
+function Base.display(x::TArray)
+    arr = x.orig_task.storage[x.ref][2]
+    @warn "display(::TArray) prints the originating task's storage, " *
+        "not the current task's storage. " *
+        "Please use show(::TArray) to display the current task's version of a TArray."
+    display(arr)
+end
```
devmotion, line 143: One should basically never define `display` but only `show(io::IO, x::TArray)` (for the compact one-line display) and `show(io::IO, ::MIME"text/plain", x::TArray)` for the more detailed multi-line representation. See https://docs.julialang.org/en/v1/manual/types/#man-custom-pretty-printing for more explanations.

src/tarray.jl
```diff
@@ -195,3 +101,193 @@ function tfill(val::Real, dim)
...
+
+Base.show(io::IO, x::TArray) = Base.show(io::IO, task_local_storage(x.ref)[2])
+
+function Base.summary(io::IO, x::TArray)
+  print(io, "Task Local Array: ")
+  summary(io, get(x))
+end
```
devmotion, line 150: The default should be fine here, I guess? According to the documentation
```
  For arrays, returns a string of size and type info, e.g. 10-element
  Array{Int64,1}.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> summary(1)
  "Int64"

  julia> summary(zeros(2))
  "2-element Array{Float64,1}"
```
So the default "2-element TArray{Float64,2}" would be consistent with Base.

src/tarray.jl
```diff
@@ -195,3 +101,193 @@ function tfill(val::Real, dim)
...
+
+Base.print_array(io::IO, x::TArray) = Base.print_array(io, get(x))
```
devmotion, line 152: Is this needed if `show` forwards everything to the underlying array?

src/tarray.jl
cpfiffer: I think in that case the issue was more because we were dispatching on `Array` instead of `AbstractArray`, but I may be misremembering.

src/tarray.jl
KDr2: Yes, you are right, I just changed the position of these lines in the commit, and now I am sure they are not necessary, I'll delete it.

src/tarray.jl
KDr2: - you are again right about the function `similar`
- and it is a good point about the method name `get`, I will change it to `_get`

src/tarray.jl
KDr2: This is also an old function which I just changed its position, I read the docs you provided and changed it to `show(io::IO, ::MIME"text/plain", x::TArray)`

src/tarray.jl
KDr2: I overwrote this just to remind the user this is a task-local array, IMHO, it's a trivial improvement, let me know if you strongly think it is not necessary  :)

src/tarray.jl
KDr2: removed.

src/tarray.jl
devmotion: I don't have a strong opinion here, I mainly thought that the type `TArray` printed in the summary already contains this information.

src/tarray.jl
KDr2: removed:
- `eltype`
- `length`
- `ndims`
- `collect`

yebai (approved):
Thanks, @KDr2 - the PR looks good to me. The performance regression between `TArray` and `Array` issue likely requires some careful optimisation but can be done in a separate PR as we discussed.

devmotion:
Maybe it would be nice to separate the implementations for different packages in different files to make the code easier to read. Probably it would be good to add more tests as well. BTW why did the PR re-add a `deps` folder?

yebai:
> Probably it would be good to add more tests as well

I should probably wait a bit longer before merging. Overall I think this PR is correct and further improvements can be done in future PRs. 


> why did the PR re-add a deps folder?

I think the script in `deps` is only for finding all the functions in standard lib that has at least one of the arguments of type `AbstractArrays`. This is for ubility purpose and is not actually used by `Libtask`. @KDr2 Maybe we can move this to `src/utilities.jl` ?

devmotion:
> I think the script in deps is only for finding all the functions in standard lib that has at least one of the arguments of type AbstractArrays. This is for ubility purpose and is not actually used by Libtask. @KDr2 Maybe we can move this to src/utilities.jl ?

Ah, I see. Maybe one could just keep it at the top level or in a separate `utils` directory? I was mainly surprised since I associate the `deps` folder with building of binary dependencies by BinaryBuilder. Similarly, I would expect all files in `src` to be included in the Julia package.

KDr2:
OK, I will move it to `utils` directory.

PR #83, opened 2021-01-07
TArray Indexing Performance

KDr2:
Trying to benchmark, profile, and improve the performance of `TArray` indexing.


test/benchmarks.jl
```diff
@@ -1,21 +1,32 @@
...
+INTENSITY = 2
+
 println("= Benchmarks on Arrays =")
-A = rand(100, 100)
-x, y =  abs.(rand(Int, 2) .% 100)
+A = rand(1000, 1000)
+x, y =  abs.(rand(Int, 2) .% 999) .+ 1
 print("indexing: ")
-@btime $A[$x, $y] + $A[$x, $y]
+@btime @rep INTENSITY $A[$x, $y]
```
devmotion, line 19: Does this interpolate the variables correctly in the `@btime` macro? For accurate benchmarking it might be good to follow the standard suggestions and benchmark proper functions with arguments being interpolated, i.e., something like `@btime f($x, $y)` (instead of `@btime f(x, y)`).

devmotion:
I assume that performance of `getindex` can be improved by adding `@propagate_inbounds` (see https://docs.julialang.org/en/v1/devdocs/boundscheck/#Propagating-inbounds).

KDr2:
> I assume that performance of `getindex` can be improved by adding `@propagate_inbounds` (see https://docs.julialang.org/en/v1/devdocs/boundscheck/#Propagating-inbounds).

It decreases about 50% time usage 👍

test/benchmarks.jl
KDr2: I think it expands correctly, I checked the macro expand results. I changed them to the function calling form, the results didn't change.

devmotion:
I am not sure if it is useful for `view` as well.

KDr2:
Results of the benchmarks:

### Time of Indexing (ns)

| Intensity | Array | TArray  |
| ----------|-------|---------|
| 1         |2.541  |424.809  |
| 2         |2.541  |814.048  |
| 3         |2.541  |1277     |
| 4         |2.544  |1711     |
| 5         |2.543  |2065     |
| 6         |2.543  |2551     |

### Time of Set-Indexing (ns)

| Intensity | Array | TArray  |
| ----------|-------|---------|
| 1         |2.876  |505.067  |
| 2         |2.881  |1029     |
| 3         |2.843  |1561     |
| 4         |2.882  |2041     |
| 5         |4.225  |2496     |
| 6         |4.225  |3037     |


### Time of Broadcasting (ms)

| Intensity | Array | TArray  |
| ----------|-------|---------|
| 1         |2.869  |6.898    |
| 2         |5.971  |13.976   |
| 3         |8.938  |21.024   |
| 4         |2.882  |2041     |
| 5         |15.225 |35.192   |
| 6         |19.283 |44.172   |

KDr2:
I did some optimization for TArray{Float64, 2} indexing: move these arrays from task local storage to a field of CTask, which makes it type stable, and here's the benchmark results:


Time of Indexing (ns)

| Intensity | Array | TArray  |optimized for Matrix{Float64} |
| ----------|-------|---------|----|
| 1         |2.541  |424.809  |34  |
| 2         |2.541  |814.048  |67  |
| 3         |2.541  |1277     |99  |
| 4         |2.544  |1711     |124 |
| 5         |2.543  |2065     |152 |
| 6         |2.543  |2551     |207 |

Time of Set-Indexing (ns)

| Intensity | Array | TArray  |optimized for Matrix{Float64}
| ----------|-------|---------|----|
| 1         |2.876  |505.067  |146 |
| 2         |2.881  |1029     |292 |
| 3         |2.843  |1561     |443 |
| 4         |2.882  |2041     |568 |
| 5         |4.225  |2496     |719 |
| 6         |4.225  |3037     |876 |


Indexing time consumption is reduced to 1/10.  The caveat is that one must create a TArray{Float64,2} is a CTask, if one tries to create a TArray{Float64,2} in an ordinary task, an error will come out.

devmotion:
It feels strange and somewhat not right to hard code a special case for `Matrix{Float64}`. Can we instead add the type of the task-local array as a type parameter to `TArray` and/or add type assertions to fix the type instability? I mean, we know the type when we construct the `TArray`, don't we?

KDr2:
> It feels strange and somewhat not right to hard code a special case for `Matrix{Float64}`. Can we instead add the type of the task-local array as a type parameter to `TArray` and/or add type assertions to fix the type instability? I mean, we know the type when we construct the `TArray`, don't we?

I tried simply adding type assertions but it didn't bring any performance improvement:
```
n, d::Array{T, N} = task_local_storage(...)
```
Maybe I didn't use it correctly?

I discussed with Hong and will try another approach: move the underlying Array to struct TArray.

KDr2:
@yebai 

Moved TArray's underlying data from task local storage to TArray.data, here's the benchmark results:

Time of Indexing (ns)

| Intensity | Array | old TArray |Data in TArray |
| ----------|-------|---------|----|
| 1         |2.541  |424.809  |57  |
| 2         |2.541  |814.048  |207 |
| 3         |2.541  |1277     |159 |
| 4         |2.544  |1711     |208 |
| 5         |2.543  |2065     |268 |
| 6         |2.543  |2551     |309 |

Time of Set-Indexing (ns)

| Intensity | Array | old TArray |Data in TArray |
| ----------|-------|---------|-----|
| 1         |2.876  |505.067  |193  |
| 2         |2.881  |1029     |383  |
| 3         |2.843  |1561     |568  |
| 4         |2.882  |2041     |777  |
| 5         |4.225  |2496     |988  |
| 6         |4.225  |3037     |1181 |


Time of Broadcasting (ms)

| Intensity | Array | Data in TArray |
| ----------|-------|--------|
| 1         |1.139  |1.160   |
| 2         |2.298  |2.299   |
| 3         |3.485  |3.483   |
| 4         |4.704  |4.687   |
| 5         |5.779  |5.832   |
| 6         |7.732  |7.780   |

src/tarray.jl
```diff
@@ -23,10 +23,15 @@ for i in 1:4 ta[i] = i end  # assign
 Array(ta)                   # convert to 4-element Array{Int64,1}: [1, 2, 3, 4]
 ```
 """
-struct TArray{T,N} <: AbstractArray{T,N}
-    ref :: Symbol  # object_id
+struct TArray{T, N} <: AbstractArray{T, N}
     orig_task :: Task
-    TArray{T,N}() where {T,N} = new(gensym(), current_task())
+    data::Dict{Task, Tuple{Int, AbstractArray{T, N}}}
```
devmotion, line 28: `data` should be concretely typed if possible, other `get(data, task)` is not inferrable.

src/tarray.jl
```diff
@@ -37,15 +42,52 @@ TArray{T,N}(::UndefInitializer, d::Vararg{<:Integer,N}) where {T,N} = TArray{T,N
...
 end
 
 TArray(x::AbstractArray) = convert(TArray, x)
 
+# TArray House-Keeper
+
+const TArrayKeeper = Vector{WeakRef}()
+keep(x::TArray) = push!(TArrayKeeper, WeakRef(x))
+function copy_tarrays(task1::Task, task2::Task)
+    deleteat!(TArrayKeeper, map(x -> x == nothing, TArrayKeeper))
```
devmotion, line 59: A bit more efficient
```suggestion
    filter!(x -> x !== nothing, TArrayKeeper)
```
(but probably still not cheap)

src/tarray.jl
```diff
@@ -37,15 +42,52 @@ TArray{T,N}(::UndefInitializer, d::Vararg{<:Integer,N}) where {T,N} = TArray{T,N
...
+    for wref in TArrayKeeper
+        ta = wref.value
+        if haskey(ta.data, task1) && !haskey(ta.data, task2)
+            ta.data[task2] = ta.data[task1]
+        end
+    end
+end
+
+# _local_storage
+_local_storage(x::TArray{T, N}) where {T, N} = x.data[current_task()]
```
devmotion, line 69:
```suggestion
_local_storage(x::TArray) = x.data[current_task()]
```

src/tarray.jl
```diff
@@ -37,15 +42,52 @@ TArray{T,N}(::UndefInitializer, d::Vararg{<:Integer,N}) where {T,N} = TArray{T,N
...
+    deleteat!(TArrayKeeper, map(x -> x == nothing, TArrayKeeper))
+    for wref in TArrayKeeper
+        ta = wref.value
+        if haskey(ta.data, task1) && !haskey(ta.data, task2)
+            ta.data[task2] = ta.data[task1]
+        end
+    end
+end
+
+# _local_storage
```
devmotion, line 68: It is a bit confusing that `_local_storage` is used both for `push!` and `get`.

src/tarray.jl
devmotion: It can be inferred currently (see above)

src/tarray.jl
```diff
@@ -37,15 +42,52 @@ TArray{T,N}(::UndefInitializer, d::Vararg{<:Integer,N}) where {T,N} = TArray{T,N
...
+_local_storage(x::TArray{T, N}) where {T, N} = x.data[current_task()]
+function _local_storage(x::TArray{T, N}, v::Tuple{Int, AbstractArray{T, N}}) where {T, N}
+    x.data[current_task()] = v
+end
+
+_get(x) = x
+function _get(x::TArray{T, N}) where {T, N}
```
devmotion, line 75:
```suggestion
function _get(x::TArray)
```

src/tarray.jl
```diff
@@ -37,15 +42,52 @@ TArray{T,N}(::UndefInitializer, d::Vararg{<:Integer,N}) where {T,N} = TArray{T,N
...
+    n, d = x.data[current_task()]
+    return d
```
devmotion, line 77: Can't be inferred either (see above)

src/tarray.jl
```diff
@@ -37,15 +42,52 @@ TArray{T,N}(::UndefInitializer, d::Vararg{<:Integer,N}) where {T,N} = TArray{T,N
...
+end
+function _get_for_write(x::TArray{T, N}) where {T, N}
```
devmotion, line 79:
```suggestion
function _get_for_write(x::TArray)
```

src/tarray.jl
```diff
@@ -121,27 +158,32 @@ function Base.convert(::Type{Array{T,N}}, x::TArray{T,N}) where {T,N}
...
     @warn "Here shows the originating task's storage, " *
         "not the current task's storage. " *
         "Please explicitly call show(::TArray) to display the current task's version of a TArray."
     show(io,  MIME("text/plain"), arr)
 end
 
-Base.show(io::IO, x::TArray) = Base.show(io::IO, task_local_storage(x.ref)[2])
+function Base.show(io::IO, x::TArray)
+    if haskey(x.data, current_task())
+        return Base.show(io::IO, _get(x))
```
devmotion, line 183:
```suggestion
        return show(io, _get(x))
```

src/tarray.jl
```diff
@@ -121,27 +158,32 @@ function Base.convert(::Type{Array{T,N}}, x::TArray{T,N}) where {T,N}
...
+    end
+    show(io, MIME("text/plain"), x)
```
devmotion, line 185: This seems conceptually wrong, it should not forward to the implementation for more verbose output. Instead one should add a custom compact output (see https://docs.julialang.org/en/v1/manual/types/#man-custom-pretty-printing-1).

src/tarray.jl
```diff
@@ -174,42 +216,22 @@ end
...
 
-function Base.pop!(x::TArray)
-    n, d = task_local_storage(x.ref)
-    cn   = n_copies()
-    newd = d
-    if cn > n
-        newd = deepcopy(d)
-        task_local_storage(x.ref, (cn, newd))
-    end
+function Base.pop!(x::TArray{T, N}) where {T, N}
```
devmotion, line 233:
```suggestion
function Base.pop!(x::TArray)
```

devmotion (commented):
It would be good to add more comments and explanations, also to motivate specific design choices. Sometimes it is a bit difficult to read and understand the implementation.

src/tarray.jl
KDr2: Indeed, I used `Array{T, N}` at the beginning but not `AbstractArray{N, T}` and the indexing performance was better than it is now, but if we use a concrete array type, we can't wrap all kinds of arrays into a `TArray`, e.g. `SubArray` returned from `view`.

Using `AbstractArray` brings a little performance decrease but not that bad I thought, and it is compatible with all kinds of arrays.

src/tarray.jl
KDr2: Done, BTW, it should be `!=` not `!==`.

> (but probably still not cheap)
This function only executes on task copying, so I think it's fine.

src/tarray.jl
KDr2: Done.

src/tarray.jl
KDr2: Sorry, this piece of code was for debugging, and I forgot to change it back, fixed now.

src/tarray.jl
devmotion: Do you get incorrect results with `!==`? Usually it is better to compare singletons with `===` and `!==`, see e.g. the discussion in https://github.com/JuliaLang/julia/issues/35585#issuecomment-619053871.

KDr2: Yeah, `x` here is a `WeakRef`, `==` and `!=` are used to check its content, that what I want here:

```
julia> struct T i end

julia> t = T(1)
T(1)

julia> r = WeakRef(t)
WeakRef(T(1))

julia> r == nothing, r === nothing
(false, false)

julia> t = 0; GC.gc();

julia> r == nothing, r === nothing
(true, false)

julia> r != nothing, r !== nothing
(false, true)
```

src/tarray.jl
devmotion: I had in mind something like
```julia
struct TArray{T,N,A<:AbstractArray{T,N}} <: AbstractArray{T,N}
    orig::Task
    data::Dict{Task,Tuple{Int,A}}
end
```
Could we do this? Or are there any problems?

src/tarray.jl
devmotion: I see, it only works when one checks the field of the `WeakRef`:
```julia
julia> r.value === nothing
true
```

devmotion: It seems in the last case everything just compiles away but I could imagine it still being slightly more performant in general:
```julia
julia> f1(x) = x == nothing
f1 (generic function with 1 method)

julia> f2(x) = x.value === nothing
f2 (generic function with 1 method)

julia> r = WeakRef(T(1))
WeakRef(T(1))

julia> @btime f1($r);
  16.485 ns (0 allocations: 0 bytes)

julia> @btime f2($r);
  1.449 ns (0 allocations: 0 bytes)
```

src/tarray.jl
KDr2: Good idea, I will have a try tomorrow.

KDr2: It seems this works very well, see the latest commit.

devmotion: Great!

KDr2:
New benchmark results after commit "underlying array type as a parametric type of TArray":

Time of Indexing (ns)

| Intensity | Array | old TArray |Data in TArray |
| ----------|-------|---------|----|
| 1         |2.541  |424.809  |21  |
| 2         |2.541  |814.048  |37  |
| 3         |2.541  |1277     |54  |
| 4         |2.544  |1711     |69  |
| 5         |2.543  |2065     |85  |
| 6         |2.543  |2551     |100 |

Time of Set-Indexing (ns)

| Intensity | Array | old TArray |Data in TArray |
| ----------|-------|---------|-----|
| 1         |2.876  |505.067  |137  |
| 2         |2.881  |1029     |276  |
| 3         |2.843  |1561     |389  |
| 4         |2.882  |2041     |550  |
| 5         |4.225  |2496     |667  |
| 6         |4.225  |3037     |785  |


Time of Broadcasting (ms)

| Intensity | Array | Data in TArray |
| ----------|-------|--------|
| 1         |2.870  |2.999   |
| 2         |5.987  |6.089   |
| 3         |8.958  |9.127   |
| 4         |12.280 |12.454  |
| 5         |15.337 |15.416  |
| 6         |19.263 |19.596  |

src/tarray.jl
```diff
@@ -96,52 +145,47 @@ Array(tz)                     # convert to 4-element Array{Float64,1}:  [9.0  9.
...
-    arr = x.orig_task.storage[x.ref][2]
+    arr = x.data[x.orig_task][2]
     @warn "Here shows the originating task's storage, " *
         "not the current task's storage. " *
         "Please explicitly call show(::TArray) to display the current task's version of a TArray."
     show(io,  MIME("text/plain"), arr)
 end
 
-Base.show(io::IO, x::TArray) = Base.show(io::IO, task_local_storage(x.ref)[2])
+Base.show(io::IO, x::TArray) = show(io::IO, _get(x))
```
devmotion, line 188: Just a minor suggestion:
```suggestion
Base.show(io::IO, x::TArray) = show(io, _get(x))
```

src/tarray.jl
```diff
@@ -259,7 +283,7 @@ Base.:*(x::AbstractArray, y::TArray) = x * _get(y) |> localize
 Base.:*(x::TArray, y::AbstractArray) = _get(x) * y |> localize
 
 # broadcast
-Base.BroadcastStyle(::Type{TArray{T, N}}) where {T, N} = Broadcast.ArrayStyle{TArray}()
+Base.BroadcastStyle(::Type{TArray{T, N, A}}) where {T, N, A} = Broadcast.ArrayStyle{TArray}()
```
devmotion, line 286: Are there some problems with dropping the type information here? It seems it is not used anyway in the implementation of `broadcasted` so it should be fine, I assume?

KDr2: I am also curious about this: if I use `::Type{TArray}` or `::Type{TArray{T, N}}`, the broadcasting benchmark will consume about 10x as much time as using `::Type{TArray{T, N, A}}`.

devmotion: Maybe it falls back to some more expensive implementation in Julia base if the type is not concrete?

devmotion: But initial question was more about `ArrayStyle{TArray}` than `::Type{TArray{T,N,A}}` though.

KDr2: Oh, `ArrayStyle{TArray}` is just a singleton used to dispatch function calls to the propriety `Broadcast.broadcasted` implementation, so I think it's OK to drop the parametric types.

devmotion: According to the [documentation](https://docs.julialang.org/en/v1/manual/interfaces/#Broadcast-Styles) for `AbstractArray`s the preferred definition is
```suggestion
Base.BroadcastStyle(::Type{<:TArray}) = Broadcast.ArrayStyle{TArray}()
```
which is basically what you had above (and therefore should give similar performance).

src/tarray.jl
```diff
@@ -259,7 +283,7 @@ Base.:*(x::AbstractArray, y::TArray) = x * _get(y) |> localize
...
 Broadcast.broadcasted(::Broadcast.ArrayStyle{TArray}, f, args...) = f.(_get.(args)...) |> localize
```
devmotion, line 287: Can we remove this method? Or, if the default fallback is not sufficient (I assume it's not), implement 
```julia
Base.similar(bc::Broadcast.Broadcasted{Broadcast.ArrayStyle{TArray}}, ::Type{ElType}) where ElType
```
instead as suggested in the [example in the documentation](https://docs.julialang.org/en/v1/manual/interfaces/#Selecting-an-appropriate-output-array)?

KDr2: I tried this:

```diff
--- a/src/tarray.jl
+++ b/src/tarray.jl
@@ -46,6 +46,7 @@ TArray{T}(::UndefInitializer, dim::NTuple{N,Int}) where {T,N} = TArray(T, dim)
 TArray{T,N}(d::Vararg{<:Integer,N}) where {T,N} = TArray(T, d)
 TArray{T,N}(::UndefInitializer, d::Vararg{<:Integer,N}) where {T,N} = TArray{T,N}(d)
 TArray{T,N}(dim::NTuple{N,Int}) where {T,N} = TArray(T, dim)
+TArray{T, N, Array{T, N}}(::UndefInitializer, dim::NTuple{N,Int}) where {T,N} = TArray(T, dim)

 function TArray(T::Type, dim)
     N_dim = length(dim)
@@ -283,8 +284,10 @@ Base.:*(x::AbstractArray, y::TArray) = x * _get(y) |> localize
 Base.:*(x::TArray, y::AbstractArray) = _get(x) * y |> localize

 # broadcast
-Base.BroadcastStyle(::Type{TArray{T, N, A}}) where {T, N, A} = Broadcast.ArrayStyle{TArray}()
-Broadcast.broadcasted(::Broadcast.ArrayStyle{TArray}, f, args...) = f.(_get.(args)...) |> localize
+Base.BroadcastStyle(::Type{TArray{T, N, A}}) where {T, N, A} = Broadcast.ArrayStyle{TArray{T, N, A}}()
+# Broadcast.broadcasted(::Broadcast.ArrayStyle{TArray}, f, args...) = f.(_get.(args)...) |> localize
+Base.similar(bc::Broadcast.Broadcasted{Broadcast.ArrayStyle{TArray{T, N, A}}}, ::Type{T}) where {T, N, A} =
+    similar(TArray{T, N, Array{T, N}}, axes(bc))

 import LinearAlgebra
 import LinearAlgebra:  \, /, inv, det, logdet, logabsdet, norm
```

It became much slower (2.8 ms -> 192 ms).

KDr2:
About the benchmarks of `getindex`, I checked generated LLVM IR, multiple get-indexing expressions in a function are optimized to one expression:
```
f(A, i)= begin
A[i]
A[i]
A[i]
A[i]
end
```
becomes

```
f(A, i) = A[i]
```

Because the compiler knows that if A is a built-in Array, only one expression is used (as the return value).

I will try to find a way to prevent this optimization.

src/tarray.jl
```diff
@@ -7,7 +7,13 @@
 
 Implementation of data structures that automatically perform copy-on-write after task copying.
 
-If current_task is an existing key in `s`, then return `s[current_task]`. Otherwise, return `s[current_task] = s[last_task]`.
+Underlying `TArray`, we have an concrete array object for each task, i,e, if you access (read
+or write) a same `TArray` is different tasks, you may get different result because you are
+eventaully accessing different arrays wrapped by the `TArray` object.
+
+Now we store the underlying arrays in the field `TArray.data`, a dict whose keys are tasks,
```
yebai, line 14:
```suggestion
More specifically, we store the underlying arrays in the field `TArray.data`, a dictionary whose keys are tasks,
```

src/tarray.jl
```diff
@@ -7,7 +7,13 @@
...
+instead of storing them in the task local storage, because the latter way can't provide
```
yebai, line 15:
```suggestion
instead of storing them in the task local storage, for performance considerations. 
```

src/tarray.jl
```diff
@@ -7,7 +7,13 @@
...
+assistance for type inference thus may bring a perfomance penalty.
```
yebai, line 16:
```suggestion
```

src/tarray.jl
```diff
@@ -23,10 +29,15 @@ for i in 1:4 ta[i] = i end  # assign
...
-struct TArray{T,N} <: AbstractArray{T,N}
-    ref :: Symbol  # object_id
+struct TArray{T, N, A <: AbstractArray{T, N}} <: AbstractArray{T, N}
     orig_task :: Task
-    TArray{T,N}() where {T,N} = new(gensym(), current_task())
+    data::Dict{Task, Tuple{Int, A}}
+    function TArray{T, N, A}() where {T, N, A <: AbstractArray{T, N}}
+        d = Dict{Task, Tuple{Int, A}}()
+        res = new(current_task(), d)
+        keep(res)
```
yebai, line 38: what does `keep` mean?

src/tarray.jl
```diff
@@ -37,15 +48,53 @@ TArray{T,N}(::UndefInitializer, d::Vararg{<:Integer,N}) where {T,N} = TArray{T,N
...
+    _local_storage(res, (n, d))
     res
 end
 
 TArray(x::AbstractArray) = convert(TArray, x)
 
+# TArray House-Keeper
+
+const TArrayKeeper = Vector{WeakRef}()
+keep(x::TArray) = push!(TArrayKeeper, WeakRef(x))
```
yebai, line 64: `keep` is a bit confusing. Maybe consider a more informative/specific function name?

src/tarray.jl
yebai: Seconded - maybe consider `get_local_storage` and `set_local_storage`?

src/tarray.jl
yebai: @devmotion Just to clarify, do you mean the types of `x::TArray` need to be parametric / more concrete?

src/tarray.jl
```diff
@@ -37,15 +48,53 @@ TArray{T,N}(::UndefInitializer, d::Vararg{<:Integer,N}) where {T,N} = TArray{T,N
...
+function _local_storage(x::TArray{T, N}, v::Tuple{Int, AbstractArray{T, N}}) where {T, N}
+    x.data[current_task()] = v
+end
+
+_get(x) = x
+function _get(x::TArray)
+    n, d = x.data[current_task()]
+    return d
+end
+function _get_for_write(x::TArray)
```
yebai, line 86: Maybe consider unifying `_get` and `_get_for_write`, e.g. by adding a flag to force deepcopy?

src/tarray.jl
yebai: @KDr2 `_get` and `_get_for_write` looks similiar to `_local_storage`. If that's true, maybe merge these functions into `_get_local_storage` / `_set_local_storage` for clarity?

src/tarray.jl
yebai: > Base.BroadcastStyle(::Type{<:TArray}) = Broadcast.ArrayStyle{TArray}()

I'd second the one from Julia documentation since it is more concise.

src/tarray.jl
yebai: thanks, @KDr2 and @devmotion!

yebai (changes requested):
thanks @KDr2 - I've added some comments below. The PR looks good in general, however, there are a few places that can be improved for clarity/performance

src/tarray.jl
devmotion: > Just to clarify, do you mean the types of x::TArray need to be parametric / more concrete?

Yes, I mean that the type of `d = last(x.data[task])` can only be inferred if `data::Dict{Task,Tuple{Int,V}}` is a field of `TArray` with concrete value type `V` that is e.g. a type parameter of `TArray`.

Just to clarify, this is not related to the type parameters `T` and `N` and corresponding where clauses in the definition of `_get`. They are not helpful for type inference.

src/tarray.jl
KDr2: Register the `TArray` object to the House-Keeper below.

KDr2:
The latest commit:
- documentation update 
- renamed function `keep` to `register_to_keeper`
- revised local storage functions: `_local_stroge`, `_get`, `_get_for_write`  to `_set_local_storage` and `_get_local_storage`
- using `Base.similar` in implementation of broadcasting (though not performant)

devmotion:
> using `Base.similar` in implementation of broadcasting (though not performant)

Why did you change it? I would be fine with the implementation of `broadcasted`, I was mainly curious if and why it is needed. I only suggested to simplify the implementation of `BroadcastStyle` slightly to
```julia
Base.BroadcastStyle(::Type{<:TArray}) = Broadcast.ArrayStyle{TArray}()
```
as shown in the documentation.

yebai (approved):
Looks good - happy to merge once the broadcasting issue raised by @devmotion is addressed!

KDr2:
> > using `Base.similar` in implementation of broadcasting (though not performant)
> 
> Why did you change it? I would be fine with the implementation of `broadcasted`, I was mainly curious if and why it is needed. I only suggested to simplify the implementation of `BroadcastStyle` slightly to
> 
> ```julia
> Base.BroadcastStyle(::Type{<:TArray}) = Broadcast.ArrayStyle{TArray}()
> ```
> 
> as shown in the documentation.

Sorry, I misunderstood, fixed now.

PR #86, opened 2021-06-13
fix DynamicPPL.jl/pull/261

KDr2:
https://github.com/TuringLang/Libtask.jl/pull/new/DynamicPPL/pull/261

The issue was caused by the usage of WeakRef to an immutable object:

An immutable object could be moved under some circumstances, but the weak ref to it won't know it's moved, it will see moving as a GC, and mark its referencing value as nothing.

I changed the struct `TArray` from immutable to mutable to avoid this false elimination, and `TArray` only has two pointer fields, so if it is mutable or immutable doesn't make much difference, I think.

### DIFF
--- a/src/tarray.jl
+++ b/src/tarray.jl
@@ -31,7 +31,7 @@ for i in 1:4 ta[i] = i end  # assign
 Array(ta)                   # convert to 4-element Array{Int64,1}: [1, 2, 3, 4]
 ```
 """
-struct TArray{T, N, A <: AbstractArray{T, N}} <: AbstractArray{T, N}
+mutable struct TArray{T, N, A <: AbstractArray{T, N}} <: AbstractArray{T, N}
     orig_task :: Task
     data::Dict{Task, Tuple{Int, A}}
     function TArray{T, N, A}() where {T, N, A <: AbstractArray{T, N}}
@@ -61,14 +61,13 @@ end
 TArray(x::AbstractArray) = convert(TArray, x)
 
 # TArray House-Keeper
-
 const TArrayKeeper = Vector{WeakRef}()
 register_to_keeper(x::TArray) = push!(TArrayKeeper, WeakRef(x))
 function copy_tarrays(task1::Task, task2::Task)
     filter!(x -> x.value !== nothing, TArrayKeeper)
     for wref in TArrayKeeper
         ta = wref.value
-        if haskey(ta.data, task1) && !haskey(ta.data, task2)
+        if ta !== nothing && haskey(ta.data, task1) && !haskey(ta.data, task2)
             ta.data[task2] = ta.data[task1]
         end
     end


KDr2:
The memory layout of `jl_task_t` changed again in the master branch, that's why the test failed on nightly build.

yebai:
> I changed the struct TArray from immutable to mutable to avoid this false elimination, and TArray only has two pointer fields, so if it is mutable or immutable doesn't make much difference, I think.

It'll be helpful to have some basic benchmarking for the change, I think.

devmotion (commented):
Can we add a test for the bug? I'll also open a separate PR with some integration tests for Turing and AdvancedPS but they don't help if Libtask is upper bounded explicitly.

KDr2:
> Can we add a test for the bug? I'll also open a separate PR with some integration tests for Turing and AdvancedPS but they don't help if Libtask is upper bounded explicitly.

I am not sure whether there's a certain way to cause an immutable object to be moved, thus we can construct a test case for this.  As how it is designed, the moving of objects is opaque to the Julia code and can't be explicitly controlled by Julia code, but I will try to construct one.

KDr2:
> It'll be helpful to have some basic benchmarking for the change, I think.

I ran the old benchmarks, here are the results:
Before:
```
= Benchmarks on Arrays =
indexing:   2.554 ns (0 allocations: 0 bytes)
set indexing:   4.416 ns (0 allocations: 0 bytes)
broadcast:   20.374 ms (12 allocations: 45.78 MiB)
= Benchmarks on TArrays =
indexing:   137.360 ns (0 allocations: 0 bytes)
set indexing:   972.640 ns (0 allocations: 0 bytes)
broadcast:   21.410 ms (78 allocations: 45.78 MiB)
```
After:
```
= Benchmarks on Arrays =
indexing:   2.556 ns (0 allocations: 0 bytes)
set indexing:   4.708 ns (0 allocations: 0 bytes)
broadcast:   21.192 ms (12 allocations: 45.78 MiB)
= Benchmarks on TArrays =
indexing:   134.313 ns (0 allocations: 0 bytes)
set indexing:   813.130 ns (0 allocations: 0 bytes)
broadcast:   21.429 ms (60 allocations: 45.78 MiB)
```

But these are mostly about data accessing, rather than the placement and moving of `TArray` objects. We can finely control the moving of immutable objects so it's hard to construct a benchmark about that. 

I will do a benchmark if I can successfully construct a test case for this issue. 

After all, I guess the move may occur when a ctask is copied, but not every time a ctask is copied. So its usage in this aspect is not frequent. Meantime, the underlying array data is not directly in the `TArray` so the placement of it doesn't change during this commit. So I think the performance affection is really trivial, either from the perspective of data locality or memory copying.

KDr2:
I found that if you copy a task a big amount of times the immutable objects in the task will be moved, and wrote a test case according to it.

It seems not impossible to benchmark it because the old version will throw an error if you run a benchmark against it.

yebai:
Thanks, @KDr2 - ready to merge this PR. Could you add the benchmark scripts to the repo (e.g. `benchmarks/scripts.jl`) as a follow-up PR?

PR #87, opened 2021-06-13
Add integration tests

devmotion:
Adds integration tests of AdvancedPS and Turing.

### DIFF
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -1,0 +1,48 @@
+name: IntegrationTest
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


KDr2:
I think the integration test with Turing is supposed to fail at this point, but it passes because of this https://github.com/TuringLang/DynamicPPL.jl/pull/261#issuecomment-859732325. Should we edit Turing's Project.toml to ensure the latest Libtask is used while the test is running?

devmotion:
I don't think this should be done. The main point is that the integration tests are allowed to fail if we make breaking changes in Libtask and/or the current version is not supported by Turing. So one should not manually modify Turing's Project.toml in Libtask but it should only be updated downstream once it is compatible.

KDr2 (approved):

PR #89, opened 2021-06-18
use memory layout info to copy task

KDr2:


src/ctask.jl
```diff
@@ -49,7 +49,23 @@ function enable_stack_copying(t::Task)
     if istaskfailed(t)
         error("only runnable or finished tasks' stack can be copied.")
     end
-    return ccall((:jl_enable_stack_copying, libtask_julia), Any, (Any,), t)::Task
+    # return ccall((:jl_enable_stack_copying, libtask_julia), Any, (Any,), t)::Task
+    copy_stack = ccall((:jl_getfield_int32_t, libtask_julia),
+                       Int32, (Any, Csize_t), t, TASK_OFFSETS[:copy_stack])
+    if copy_stack == 0
+        ccall((:jl_setfield_int32_t, libtask_julia),
```
yebai, line 56: Maybe consider rewrapping the repetative `getfield/setfield` code block into a `setfield/getfield` function, e.g. 

```julia
function setfield(t, fieldname, value)
        ccall((:jl_setfield_int32_t, libtask_julia),
              Any, (Any, Csize_t, Int32),
              t, TASK_OFFSETS[:copy_stack], 1)
end
```

src/ctask.jl
```diff
@@ -101,7 +117,48 @@ function Base.copy(ctask::CTask)
         error("only runnable or finished tasks can be copied.")
     end
 
-    newtask = ccall((:jl_clone_task, libtask_julia), Any, (Any,), task)::Task
+    # memory copy
+    # newtask = ccall((:jl_clone_task, libtask_julia), Any, (Any,), task)::Task
+    newtask = ccall((:jl_clone_task_opaque, libtask_julia),
+                    Any, (Any, Csize_t), task, TASK_OFFSETS[:END])::Task
+    haskey(TASK_OFFSETS, :exception) && ccall(
```
yebai, line 124: Similarly, please consider replacing the `ccall` code block with a more readable `setfield/getfield` function.

yebai (changes requested):
thanks, @KDr2 - looks good! I left some minor comments below. 

Also, I would suggest that we keep the `memlayout.cpp` file under `Libtask_jll`. Assuming that we need to re-run `memlayout` inside a Github workflow,  in order to get the member layout for a new Julia release, we can import `Libtask_jll`. However, this step can be done after the code becomes more mature to avoid frequent changes to `Libtask_jll` during the development phase.

src/ctask.jl
KDr2: Done.

KDr2:
Now, it's ready to be reviewed and merged.

yebai (approved):
Looks good to me overall - one minor comment: maybe create a subfolder in `workflows` for `tasklayout` and store the Julia script and cpp source file inside `tasklayout`? 

@devmotion This PR is attempting to crate a more maintainable version of `Libtask` by doing most of the task struct manipulation operations on the Julia side, and minimise the operations on the C side (i.e. in `Libtask_jll`). The solution here is not yet perfect, but operational, and already more generic than the current mechanism (e.g. it works correctly on Julia `v.1.5.4` and `v1.5.3`). Can you take a look and see whether you have any comments?

KDr2:
> maybe create a subfolder in `workflows` for `tasklayout` and store the Julia script and cpp source file inside `tasklayout`?

Done.

PR #92, opened 2021-09-07
[WIP] Julia 1.7 Compatible

KDr2:


KDr2:
With this PR (https://github.com/JuliaPackaging/Yggdrasil/pull/3597), most test cases could pass under Julia 1.7.0 now, but there are still some random failures and hang whose root cause seems to be something other than task migrating.

@yebai.

ParadaCarleton:
> With this PR ([JuliaPackaging/Yggdrasil#3597](https://github.com/JuliaPackaging/Yggdrasil/pull/3597)), most test cases could pass under Julia 1.7.0 now, but there are still some random failures and hang whose root cause seems to be something other than task migrating.
> 
> @yebai.

Tests only seem to be failing on nightly, but I don't see where the tests for 1.7 are?

(User on the Discourse is having trouble running Turing+apparently can't downgrade to 1.6.)

yebai:
Unfortunately this PR won’t fix the issue for Julia 1.7. @KDr2 is working on a prototype which is much more robust than the current Libtask, but will likely take a bit more time.

yebai:
Closed in favour of #93

PR #100, opened 2021-12-15
Tape instruction

KDr2:
Try to implement https://github.com/TuringLang/Libtask.jl/issues/99.

PR #103, opened 2022-01-05
backport basic refactor from PR#100

KDr2:
Besides adding `TapeInstruction`, I did a little code refactor in https://github.com/TuringLang/Libtask.jl/pull/100, and I'd like to move that refactor onto master branch before doing benchmark and performance optimization.

### DIFF
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -21,6 +21,7 @@ const NULL_TAPE = Tape()
 
 function setowner!(tape::Tape, owner)
     tape.owner = owner
+    return tape
 end
 
 mutable struct Box{T}
@@ -69,10 +70,11 @@ function (instr::Instruction{F})() where F
     instr.output.val = output
 end
 
-function increase_counter(t::Tape)
+function increase_counter!(t::Tape)
     t.counter > length(t) && return
     # instr = t[t.counter]
     t.counter += 1
+    return t
 end
 
 function run(tape::Tape, args...)
@@ -82,7 +84,7 @@ function run(tape::Tape, args...)
     end
     for instruction in tape
         instruction()
-        increase_counter(tape)
+        increase_counter!(tape)
     end
 end
 
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -1,7 +1,7 @@
 abstract type AbstractInstruction end
 
 mutable struct Tape
-    tape::Vector{AbstractInstruction}
+    tape::Vector{<:AbstractInstruction}
     counter::Int
     owner
 end
@@ -31,8 +31,6 @@ val(x) = x
 val(x::Box) = x.val
 box(x) = Box(x)
 box(x::Box) = x
-any_box(x) = Box{Any}(x)
-any_box(x::Box) = Box{Any}(x.val)
 
 gettape(x) = nothing
 gettape(x::Instruction) = x.tape
@@ -88,16 +86,8 @@ function run_and_record!(tape::Tape, f, args...)
         box(f(map(val, args)...))
     catch e
         @warn e
-        any_box(nothing)
+        Box{Any}(nothing)
     end
-    ins = Instruction(f, args, output, tape)
-    push!(tape, ins)
-    return output
-end
-
-function dry_record!(tape::Tape, f, args...)
-    # We don't know the type of box.val now, so we use Box{Any}
-    output = any_box(nothing)
     ins = Instruction(f, args, output, tape)
     push!(tape, ins)
     return output
@@ -205,19 +195,6 @@ function (tf::TapedFunction)(args...)
     # TODO: use cache
     run(tf.tape, args...)
     return result(tf.tape)
-end
-
-function dry_run(tf::TapedFunction)
-    isempty(tf.tape) || (return tf)
-    @assert tf.arity >= 0 "TapedFunction need a fixed arity to dry run."
-    args = fill(nothing, tf.arity)
-    ir = IRTools.@code_ir tf.func(args...)
-    ir = intercept(ir; recorder=:dry_record!)
-    tape = IRTools.evalir(ir, tf.func, args...)
-    tf.ir = ir
-    tf.tape = tape
-    tape.owner = tf
-    return tf
 end
 
 function Base.show(io::IO, tf::TapedFunction)
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -69,6 +69,12 @@ function (instr::Instruction{F})() where F
     instr.output.val = output
 end
 
+function increase_counter(t::Tape)
+    t.counter > length(t) && return
+    # instr = t[t.counter]
+    t.counter += 1
+end
+
 function run(tape::Tape, args...)
     if length(args) > 0
         input = map(box, args)
@@ -76,7 +82,7 @@ function run(tape::Tape, args...)
     end
     for instruction in tape
         instruction()
-        tape.counter += 1
+        increase_counter(tape)
     end
 end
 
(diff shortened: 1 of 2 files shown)


KDr2:
It mainly did two things:
- Moved instruction counter from task to tape.
- Optimized the code structure of task copying.

src/tapedfunction.jl
```diff
@@ -1,30 +1,38 @@
...
-    input::Tuple
-    output
-    tape
-end
-
+abstract type AbstractInstruction end
 
 mutable struct Tape
-    tape::Vector{Instruction}
+    tape::Vector{AbstractInstruction}
```
yebai, line 4:
```suggestion
    tape::Vector{<:AbstractInstruction}
```

yebai: Using abstract types in structs/containers have performance penalties.

src/tapedfunction.jl
```diff
@@ -1,30 +1,38 @@
...
 mutable struct Box{T}
     val::T
 end
 
 val(x) = x
 val(x::Box) = x.val
 box(x) = Box(x)
+box(x::Box) = x
 any_box(x) = Box{Any}(x)
+any_box(x::Box) = Box{Any}(x.val)
```
yebai, line 35: I suggest that we use `Box{Any}` to replace all `any_box` use cases and remove `any_box` to reduce redundancy.

src/tapedfunction.jl
```diff
@@ -64,10 +72,13 @@ function (instr::Instruction{F})() where F
...
 function run(tape::Tape, args...)
-    input = map(box, args)
-    tape[1].input = input
+    if length(args) > 0
+        input = map(box, args)
+        tape[1].input = input
+    end
     for instruction in tape
         instruction()
+        tape.counter += 1
```
yebai, line 81:
```suggestion
       increase_counter(tape)
```

src/tapedtask.jl
```diff
@@ -186,18 +195,21 @@ function copy_box(old_box::Box{T}, roster::Dict{UInt64, Any}) where T
...
+    end
+    output = copy_box(x.output, roster)
+    Instruction(x.fun, input, output, on_tape)
+end
+
+function Base.copy(t::Tape, roster::Dict{UInt64, Any})
     old_data = t.tape
-    new_data = Vector{Instruction}()
-    new_tape = Tape(new_data, t.owner)
+    new_data = Vector{AbstractInstruction}()
```
yebai, line 208: Consider avoiding using container types with abstract types to improve type stability.

yebai (changes requested):

yebai:
@phipsgabler @devmotion we are now entering the performance and design optimisation and phase of the tape mechanism since the functionality has been proven correct. Please comment on any issues so that we can gradually build a lightweight, performant, maintainable tape implementation. The tape implementation will be moved into `Tapir` later when it is more mature.

src/tapedfunction.jl
KDr2: updated.

src/tapedfunction.jl
KDr2: updated.

src/tapedfunction.jl
KDr2: updated.

src/tapedtask.jl
KDr2: I will benchmark this and try to figure out a way to avoid this in the performance task.

yebai (approved):

src/tapedfunction.jl
```diff
@@ -1,30 +1,36 @@
...
 mutable struct Tape
-    tape::Vector{Instruction}
+    tape::Vector{<:AbstractInstruction}
+    counter::Int
     owner
 end
 
-Tape() = Tape(Vector{Instruction}(), nothing)
-Tape(owner) = Tape(Vector{Instruction}(), owner)
+mutable struct Instruction{F} <: AbstractInstruction
```
phipsgabler, line 9: Why does this need to be mutable? Is `owner` set after construction in some cases?

src/tapedfunction.jl
```diff
@@ -63,11 +69,20 @@ function (instr::Instruction{F})() where F
     instr.output.val = output
 end
 
+function increase_counter(t::Tape)
+    t.counter > length(t) && return
+    # instr = t[t.counter]
+    t.counter += 1
+end
```
phipsgabler, line 76:
```suggestion
function increase_counter!(t::Tape)
    t.counter > length(t) && return
    # instr = t[t.counter]
    t.counter += 1
    return t
end
```

src/tapedfunction.jl
```diff
@@ -63,11 +69,20 @@ function (instr::Instruction{F})() where F
...
-    tape[1].input = input
+    if length(args) > 0
+        input = map(box, args)
+        tape[1].input = input
+    end
     for instruction in tape
         instruction()
+        increase_counter(tape)
     end
 end
```
phipsgabler, line 87: Intuitively, I'd have expected this to also return the result of the function. Or is that extracted in another place?

src/tapedfunction.jl
```diff
@@ -1,30 +1,36 @@
...
+
+Tape() = Tape(Vector{AbstractInstruction}(), 1, nothing)
+Tape(owner) = Tape(Vector{AbstractInstruction}(), 1, owner)
 MacroTools.@forward Tape.tape Base.iterate, Base.length
 MacroTools.@forward Tape.tape Base.push!, Base.getindex, Base.lastindex
 const NULL_TAPE = Tape()
 
+function setowner!(tape::Tape, owner)
+    tape.owner = owner
+end
```
phipsgabler, line 24:
```suggestion
function setowner!(tape::Tape, owner)
    tape.owner = owner
    return tape
end
```

src/tapedfunction.jl
```diff
@@ -188,27 +195,14 @@ function (tf::TapedFunction)(args...)
...
-    ir = intercept(ir; recorder=:dry_record!)
-    tape = IRTools.evalir(ir, tf.func, args...)
-    tf.ir = ir
-    tf.tape = tape
-    tape.owner = tf
-    return tf
-end
-
 function Base.show(io::IO, tf::TapedFunction)
     buf = IOBuffer()
```
phipsgabler, line 207: Why the extra buffer? You could directly `println` to `io`.

src/tapedtask.jl
```diff
@@ -1,33 +1,32 @@
 struct TapedTaskException
     exc
```
phipsgabler, line 2:
```suggestion
    exc::Exception
```

src/tapedfunction.jl
KDr2: The result is stored in `Instruction.output`.

src/tapedfunction.jl
KDr2: Here I use an extra buffer, because that there are too many calls to `println` with `io` (which is `stdout` in most situations). And between these calls, the current task may be switched off, then Juia does some `print` in another task, and then switched on. In that case, the output will not be continuous and that makes it annoying to read.

src/tapedtask.jl
KDr2: updated.

src/tapedfunction.jl
KDr2: updated.

src/tapedfunction.jl
KDr2: updated.

src/tapedfunction.jl
KDr2: Yes.

src/tapedfunction.jl
phipsgabler: Good reason. Then put a comment there to remind future readers why this should not be refactored.

PR #104, opened 2022-01-07
Performance and Benchmarks

KDr2:


KDr2:
It's interesting, in the last commit I found `step_in` a tape is faster than running a model directly, maybe I missed something?
```
$ julia --project perf/p0.jl
  2.286 μs (38 allocations: 1.95 KiB)
  97.440 ns (1 allocation: 48 bytes)
  440.273 μs (48 allocations: 2.56 KiB)
  969.364 ns (4 allocations: 288 bytes)
```

yebai:
It is possible that `step_in` the tape executes faster than the original function `f(args...)` since the tape specialises more (e.g. removing control flows, caching input/output arguments). Note that the total runtime (of Turing inference algorithm) also depends on the `CTask` constructor:

```julia
julia> @btime t = Libtask.CTask(f, args...);
  258.923 ms (766345 allocations: 43.38 MiB)

julia> @btime Libtask.step_in(t.tf.tape, args)
  95.054 ns (1 allocation: 48 bytes)

julia> @btime f(args...)
  2.549 μs (38 allocations: 1.95 KiB)
(2.0, VarInfo (2 variables (μ, σ), dimension 2; logp: -1.2750123006e7))

```

So it appears that a lot of time is spent on repetitively constructing `CTask`. Maybe we can speed this up by resuing tapes?

KDr2:
### Without Cache:
```
$ julia --project perf/p0.jl
"Directly call..." = "Directly call..."
  2.233 μs (38 allocations: 1.95 KiB)
"CTask construction..." = "CTask construction..."
  410.719 ms (974878 allocations: 59.77 MiB)
"Step in a tape..." = "Step in a tape..."
  90.543 ns (1 allocation: 48 bytes)
"Directly call..." = "Directly call..."
  422.273 μs (48 allocations: 2.56 KiB)
"CTask construction..." = "CTask construction..."
  416.812 ms (974908 allocations: 59.77 MiB)
"Step in a tape..." = "Step in a tape..."
  923.559 ns (4 allocations: 288 bytes)
```


### With IR and Tape Cache:
```
$ julia --project perf/p0.jl
"Directly call..." = "Directly call..."
  2.117 μs (38 allocations: 1.95 KiB)
"CTask construction..." = "CTask construction..."
  99.222 μs (489 allocations: 22.02 KiB)
"Step in a tape..." = "Step in a tape..."
  87.400 ns (1 allocation: 48 bytes)
"Directly call..." = "Directly call..."
  417.133 μs (48 allocations: 2.56 KiB)
"CTask construction..." = "CTask construction..."
  103.745 μs (495 allocations: 22.48 KiB)
"Step in a tape..." = "Step in a tape..."
  924.314 ns (4 allocations: 288 bytes)
```

KDr2:
In spite of numeric test failures and a few errors, unit tests finished in about 2 hours on my machine:

```
real    130m46.448s
user    126m4.793s
sys     5m25.685s
```

src/tapedtask.jl
```diff
@@ -16,9 +16,19 @@ struct TapedTask
     end
 end
 
+const TRCache = LRU{Any, Any}(maxsize=10)
+
 function TapedTask(tf::TapedFunction, args...)
     tf.owner != nothing && error("TapedFunction is owned to another task.")
```
yebai, line 22:
```suggestion
    tf.owner != nothing && error("TapedFunction is owned by another task.")
```

src/tapedtask.jl
```diff
@@ -16,9 +16,19 @@ struct TapedTask
...
-    isempty(tf.tape) && tf(args...)
+    if isempty(tf.tape)
+        if haskey(TRCache, tf.func)
+            ir, tape = TRCache[tf.func]
+            reset!(tf, ir, copy(tape, Dict{UInt64, Any}(); partial=false))
```
yebai, line 26: We need to update the `args` stored in `tf::Tape` when using cached tape.

yebai (changes requested):

Project.toml
```diff
@@ -6,15 +6,20 @@ repo = "https://github.com/TuringLang/Libtask.jl.git"
...
 [deps]
+AdvancedPS = "576499cb-2369-40b2-a588-c64705576edc"
+BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+DynamicPPL = "366bfd00-2699-11ea-058f-f148b4cae6d8"
 IRTools = "7869d1d1-7146-5819-86e3-90919afe41df"
+LRUCache = "8ac3fa9e-de4c-5943-b1dc-09c6b5f20637"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Turing = "fce5fe82-541a-59a6-adf8-730c64b5f9a0"
```
yebai, line 17: Consider moving these additional dependencies into `perf/Project.toml`

src/tapedtask.jl
```diff
@@ -16,9 +16,19 @@ struct TapedTask
...
 end
 
+const TRCache = LRU{Any, Any}(maxsize=10)
+
 function TapedTask(tf::TapedFunction, args...)
     tf.owner != nothing && error("TapedFunction is owned to another task.")
-    isempty(tf.tape) && tf(args...)
+    if isempty(tf.tape)
+        if haskey(TRCache, tf.func)
+            ir, tape = TRCache[tf.func]
```
yebai, line 25: We need to cache by call signature, i.e. `TRCache[(tf.func, typeof.(args)...)]` to avoid incorrect function dispatching.

perf/p0.jl
```diff
@@ -0,0 +1,47 @@
+# ]add  Turing#hg/new-libtask2
+
+using Libtask
+using Turing, DynamicPPL, AdvancedPS
+using BenchmarkTools
+
+@model gdemo(x, y) = begin
```
devmotion, line 7: DynamicPPL contains a set of test models, maybe they can be used?

src/tapedtask.jl
```diff
@@ -16,9 +16,22 @@ struct TapedTask
     end
 end
 
+const TRCache = LRU{Any, Any}(maxsize=10)
+
 function TapedTask(tf::TapedFunction, args...)
-    tf.owner != nothing && error("TapedFunction is owned to another task.")
-    isempty(tf.tape) && tf(args...)
+    tf.owner != nothing && error("TapedFunction is owned by another task.")
```
devmotion, line 22:
```suggestion
    tf.owner !== nothing && error("TapedFunction is owned by another task.")
```

perf/p0.jl
yebai: @devmotion  DynamicPPL test [models](https://github.com/TuringLang/DynamicPPL.jl/blob/master/src/test_utils.jl) doesn't currently include `gdemo`. On the other hand, it is useful to have a complete MWE here to test the efficiency and correctness of Libtask.

perf/src/LibtaskPerf.jl
```diff
@@ -0,0 +1,2 @@
+module LibtaskPerf
```
yebai, line 1: @KDr2 Why do we need this empty package? If it is for `Project.toml`, maybe we can explicitly call `activate perf` to avoid this dummy package.

yebai (approved):
I can confirm that the tests now run correctly - we can rerun the Turing CI once this PR is merged. Fingers crossed!

perf/src/LibtaskPerf.jl
KDr2: Removed  :D

perf/p2.jl
```diff
@@ -0,0 +1,63 @@
+using Turing, Test, AbstractMCMC, DynamicPPL, Random, Turing.RandomMeasures, Libtask
```
yebai, line 1: @KDr2 Maybe write a `perf/run.jl` script to run all these examples on CI. They also serve as additional integration tests, apart from benchmarking purposes.

src/tapedfunction.jl
```diff
@@ -13,6 +13,12 @@ mutable struct Instruction{F} <: AbstractInstruction
     tape::Tape
 end
 
+mutable struct NewInstruction <: AbstractInstruction
```
yebai, line 16: Can you add some docstring/comment explaining why a special `NewInstruction` is needed?

src/tapedfunction.jl
```diff
@@ -66,10 +79,28 @@ function Base.show(io::IO, tp::Tape)
 end
 
 function (instr::Instruction{F})() where F
-    output = instr.fun(map(val, instr.input)...)
-    instr.output.val = output
+    try
```
yebai, line 82:
```suggestion
    # Catch run-time exceptions / errors.
    try
```

src/tapedfunction.jl
```diff
@@ -66,10 +79,28 @@ function Base.show(io::IO, tp::Tape)
...
+    catch e
+        println(e, catch_backtrace());
+        rethrow(e);
+    end
+end
+
+
+function (instr::NewInstruction)()
+    try
+        expr = Expr(:new, map(val, instr.input)...)
```
yebai, line 94: This is probably going to cause performance issues as we repeatedly parse and eval AST expressions at run-time. Maybe re-write as a generated function?

src/tapedfunction.jl
```diff
@@ -101,6 +132,19 @@ function run_and_record!(tape::Tape, f, args...)
...
 
+function run_and_record_new!(tape::Tape, args...)
+    output = try
+        expr = Expr(:new, map(val, args)...)
+        box(eval(expr))
+    catch e
+        @warn e
+        Box{Any}(nothing)
+    end
+    ins = NewInstruction(args, output, tape)
```
yebai, line 143: Similiar, since we know the type of `output`, can we replace the new operator with an explicit constructor (e.g. `Array{Int}(3)`).

KDr2: Unfortunately we can't do that, if there's a `new` operator, it indicates the default constructor we need is hidden by an inner constructor, use a `new` expression to access it is the only way I found.

src/tapedfunction.jl
KDr2: updated.

src/tapedfunction.jl
KDr2: updated.

src/tapedfunction.jl
rikhuijzer: Maybe I misunderstand, but the `NewInstruction` is an intermediate struct to create an `Instruction`? Just to be sure that the default way of handling this isn't overlooked, this is how structs are normally constructed when an inner constructor hides the default constructor:

```julia
struct A
    b::String

    function A(b::Float64)
        return string(b)
    end
end

A(b::Int) = A(Float64(b))

A(1.0) # "1.0"

A(1) # "1.0"
```

Apologies if I misunderstood the problem; feel free to ignore my comment then

KDr2: @rikhuijzer you are right, this is exactly the problem, we cant create an A with `A("1")`, but we can do `eval(Expr(:new, A, "string"))`.

KDr2:
This PR is ready to merge. @yebai

perf/p2.jl
KDr2: Done.

yebai (approved):

PR #106, opened 2022-01-22
Add refs to integration tests

rikhuijzer:
This way, the tests are actually executed instead of skipped in the `try catch` block due to incompatible versions. But, I also understand now that it's hard to do integration tests when changes need to be made at two repositories. Anyway, adding the integration tests here is nice because it's now easier to see what needs to be done and for the reviewer to see what the current status is.

### DIFF
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         package:
-          - {user: TuringLang, repo: AdvancedPS.jl, ref: taped-libtask}
+          - {user: TuringLang, repo: AdvancedPS.jl}
           - {user: TuringLang, repo: Turing.jl, ref: hg/new-libtask2}
 
     steps:
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -14,8 +14,8 @@ jobs:
       fail-fast: false
       matrix:
         package:
-          - {user: TuringLang, repo: AdvancedPS.jl}
-          - {user: TuringLang, repo: Turing.jl}
+          - {user: TuringLang, repo: AdvancedPS.jl, ref: taped-libtask}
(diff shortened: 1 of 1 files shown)


cpfiffer (approved):

devmotion:
Maybe open an issue that reminds us to revert this change once these packages are updated?

.github/workflows/IntegrationTest.yml
```diff
@@ -14,8 +14,8 @@ jobs:
       fail-fast: false
       matrix:
         package:
-          - {user: TuringLang, repo: AdvancedPS.jl}
-          - {user: TuringLang, repo: Turing.jl}
+          - {user: TuringLang, repo: AdvancedPS.jl, ref: taped-libtask}
```
yebai, line 17: `AdvancedPS`‘s new releases is compatible with new `Libtask`

```suggestion
          - {user: TuringLang, repo: AdvancedPS.jl}
```

yebai (approved):

yebai:
Thanks @rikhuijzer!

.github/workflows/IntegrationTest.yml
rikhuijzer: Not according to the logs of the AdvancedPS integration test run:
```
   Resolving package versions...
┌ Info: Not compatible with this release. No problem.
└   exception = empty intersection between Libtask@0.6.4 and project compatibility 0.5.3-0.5
```

yebai: It’s a bit strange - `AdvancedPS` @ v.3.2 support new versions of Libtask. In order to support the most recent Libtask release, I just updated compability bounds in AdvancedPS and made a new release. See https://github.com/JuliaRegistries/General/pull/53029

yebai: I looked into the CI message - it seems Integration test is using the `master` branch instead of the most recent release of `AdvancedPS` (i.e. v0.3.2). We currently have some breaking changes in `master` so the most recent release of AdvancedPS is based on the `releases` branch.

Issue #96, opened 2021-12-12
Implement support for stochastic control flow (e.g. if and unbounded loop)


yebai:
Related https://github.com/TuringLang/AdvancedPS.jl/issues/37

KDr2:
Is it fine to close this issue and https://github.com/TuringLang/AdvancedPS.jl/issues/37 now?

yebai:
@KDr2 there is still one broken test for constrol flows, can you take a look? 

https://github.com/TuringLang/Turing.jl/pull/1766/files#diff-995baae0024148e101534255397cc34116f06aabbba2efc91d44e59bddb88b48R55

KDr2:
I haven't found the cause yet. Does this run well if it is called directly without Libtask.CTask?

The code below may contain some clues useful:

```
62:
  %183 = Core.isdefined(%6, :contents)
  br 63 unless %183
  br 64
63:
  br 64
64:
  %184 = Core.getfield(%6, :contents)
  %185 = Base.getindex(%184, %22)
  %186 = Base.getindex(%144, %185)  # !!! <-------- Here, %144 is a 1-element vector, %185 is 2
  %187 = Main.Normal(%186, 1.0)
  %188 = (DynamicPPL.check_tilde_rhs)(%187)
  %189 = (Base.maybeview)(%21, %22)
  %190 = (DynamicPPL.tilde_observe!!)(%4, %188, %189, %150, %145)
  %191 = Base.indexed_iterate(%190, 1)
  %192 = Core.getfield(%191, 1)
  %193 = Core.getfield(%191, 2)
  %194 = Base.indexed_iterate(%190, 2, %193)
  %195 = Core.getfield(%194, 1)
  br 65 (%195, %21)
65: (%196, %197)
  %198 = Base.iterate(%14, %23)
  %199 = %198 === nothing
  %200 = Base.not_int(%199)
  br 66 (%196) unless %200
  br 2 (%198, %196, %144, %197)
66: (%201)
  %202 = Core.tuple(nothing, %201)
  return %202

```

yebai:
@KDr2 I did some further debugging. The bug seems caused by `TArray`. Here are two examples using `TArray` and `Array` (deep copy enabled) respectively, the one with `Array` runs correctly. 


<details>
  <summary>`Array` with deep copy enabled for `TapedFunction`  </summary>

```julia
julia> using Turing, Libtask, Random, Test, DynamicPPL, Turing.RandomMeasures

[...]

PR #107, opened 2022-01-23
Control flow support

KDr2:
Implement https://github.com/TuringLang/Libtask.jl/issues/96.

### DIFF
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -79,7 +79,7 @@ function reset!(tf::TapedFunction, ir::IRTools.IR, tape::RawTape)
     tf.tape = tape
     blk_map = Dict{Int, Int}()
 
-    for (i, ins) in tf.tape
+    for (i, ins) in enumerate(tf.tape)
         isa(ins, BlockInstruction) || continue
         blk_map[ins.id] = i
     end
@@ -159,7 +159,8 @@ end
 function (instr::Instruction{F})() where F
     # catch run-time exceptions / errors.
     try
-        output = instr.func(map(val, instr.input)...)
+        func = val(instr.func)
+        output = func(map(val, instr.input)...)
         instr.output.val = output
         instr.tape.counter += 1
     catch e
@@ -225,7 +226,7 @@ end
 
 function translate!(taped::Taped, ir::IRTools.IR)
     tape = taped.tape
-    boxes = Dict{Any, Box{Any}}()
+    boxes = Dict{IRTools.Variable, Box{Any}}()
     _box = (x) -> arg_boxer(x, boxes)
     for (blk_id, blk) in enumerate(IRTools.blocks(ir))
         # blocks
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -248,11 +248,11 @@ function translate!(taped::Taped, ir::IRTools.IR)
                 ins = Instruction(_new, args |> Tuple, _box(x), taped)
                 push!(tape, ins)
             else
-                try
+                if isa(st.expr, Symbol) || isa(st.expr, GlobalRef)
                     v = eval(st.expr)
                     ins = Instruction(identity, (v,), _box(x), taped)
                     push!(tape, ins)
-                catch
+                else
                     @warn "Unknown IR code: " st
                 end
             end
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -1,3 +1,7 @@
+mutable struct Box{T}
+    val::T
+end
+
 abstract type AbstractInstruction end
 abstract type Taped end
 const RawTape = Vector{AbstractInstruction}
@@ -10,7 +14,7 @@ An `Instruction` stands for a function call
 mutable struct Instruction{F, T<:Taped} <: AbstractInstruction
     func::F
     input::Tuple
-    output
+    output::Box
     tape::T
 end
 
@@ -21,7 +25,7 @@ mutable struct BlockInstruction{T<:Taped} <: AbstractInstruction
 end
 
 mutable struct BranchInstruction{T<:Taped} <: AbstractInstruction
-    condition::Any
+    condition::Union{Bool, Box{Any}, Nothing}
     block::Int
     args::Vector
     tape::T
@@ -45,10 +49,6 @@ mutable struct TapedFunction{F} <: Taped
         new{F}(f, arity, nothing, RawTape(), 1,
                Dict{Int, Int}(), nothing, nothing)
     end
-end
-
-mutable struct Box{T}
-    val::T
 end
 
 ## methods for Box
@@ -248,9 +248,13 @@ function translate!(taped::Taped, ir::IRTools.IR)
                 ins = Instruction(_new, args |> Tuple, _box(x), taped)
                 push!(tape, ins)
             else
-                @warn "Unknown IR code: " st
-                ins = Instruction(identity, (eval(st.expr),), _box(x), taped)
-                push!(tape, ins)
+                try
+                    v = eval(st.expr)
+                    ins = Instruction(identity, (v,), _box(x), taped)
+                    push!(tape, ins)
+                catch
+                    @warn "Unknown IR code: " st
+                end
             end
         end
 
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -36,6 +36,8 @@ mutable struct ReturnInstruction{TA, T<:Taped} <: AbstractInstruction
     tape::T
 end
 
+const TRCache = LRU{Any, Any}(maxsize=10)
+
 mutable struct TapedFunction{F} <: Taped
     func::F # maybe a function, a constructor, or a callable obejct
     arity::Int
@@ -46,19 +48,30 @@ mutable struct TapedFunction{F} <: Taped
     block_map::Dict{Int, Int}
     retval
     owner
-    function TapedFunction(f::F, args...; init=true) where {F}
+
+    function TapedFunction(f::F, args...; cache=false, init=true) where {F}
+        cache_key = (f, typeof.(args)...)
+
+        if cache && haskey(TRCache, cache_key) # use cache
+            cached_tf = TRCache[cache_key]
+            tf = copy(cached_tf)
+            tf.counter = 1
+            return tf
+        end
+
         tf = new{F}() # leave some fields to be undef
         tf.func = f
         tf.tape = RawTape()
         tf.counter = 1
         tf.block_map = Dict{Int, Int}()
 
-        # init
-        if init
+        if init # init
             tf.arity = length(args)
             ir = IRTools.@code_ir tf.func(args...)
             tf.ir = ir
             translate!(tf, ir)
+            # set cache
+            TRCache[cache_key] = tf
         end
         return tf
     end
@@ -337,7 +350,8 @@ function Base.copy(old_tape::RawTape, on_tape::Taped, roster::Dict{UInt64, Any})
 end
 
 function Base.copy(tf::TapedFunction)
-    new_tf = TapedFunction(tf.func; init=false)
+    # create a new uninitialized TapedFunction
+    new_tf = TapedFunction(tf.func; cache=false, init=false)
     new_tf.block_map = tf.block_map
     new_tf.ir = tf.ir
     roster = Dict{UInt64, Any}()
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -11,28 +11,28 @@ const RawTape = Vector{AbstractInstruction}
 
 An `Instruction` stands for a function call
 """
-mutable struct Instruction{F, T<:Taped} <: AbstractInstruction
+mutable struct Instruction{F, TO, T<:Taped} <: AbstractInstruction
     func::F
     input::Tuple
-    output::Box
+    output::Box{TO}
     tape::T
 end
 
-mutable struct BlockInstruction{T<:Taped} <: AbstractInstruction
-    id::Int
-    args::Vector
+mutable struct BlockInstruction{TA, T<:Taped} <: AbstractInstruction
+    block_id::Int
+    args::Vector{TA}
     tape::T
 end
 
-mutable struct BranchInstruction{T<:Taped} <: AbstractInstruction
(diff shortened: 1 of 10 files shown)


src/tapedfunction.jl
```diff
@@ -14,15 +14,36 @@ mutable struct Instruction{F, T<:Taped} <: AbstractInstruction
...
 end
 
+mutable struct BlockInstruction{T<:Taped} <: AbstractInstruction
+    id::Int
+    args::Vector
+    tape::T
+end
+
+mutable struct BranchInstruction{T<:Taped} <: AbstractInstruction
+    condition::Any
```
phipsgabler, line 24: I think this can only be a variable id or `nothing`.

src/tapedfunction.jl
```diff
@@ -14,15 +14,36 @@ mutable struct Instruction{F, T<:Taped} <: AbstractInstruction
...
+    tape::T
+end
+
 mutable struct TapedFunction{F} <: Taped
     func::F # maybe a function or a callable obejct
     arity::Int
     ir::Union{Nothing, IRTools.IR}
     tape::RawTape
     counter::Int
+    block_map::Dict{Int, Int}
```
phipsgabler, line 41: Just trying to understand the representation...

You are storing the block structure by saving all instructions linearly, and then additionally in `block_map` the mapping from block number to the id of the "block instruction" on top of a block, do I get this right? So

```
1 ─ %1 = x > 1
└──      goto #3 if not %1
2 ─ %3 = Main.rand()
└──      return %3
3 ─      return 2
```

would become, e.g.

```
1 => Block(1, x)
2 => %1 = x > 1
3 => goto 3 if not %1
4 => Block(2)
5 => %3 = Main.rand()
6 => return %3
```

with `block_map = Dict(1 => 1, 2 => 4)`?

KDr2: Yes, you are right. We only have the id of a `Block` in the `BranchInstruction`, but we don't its index (where is it on the tape). With this map, we can easily find where a Branch wants to go without iterating the whole tape.

src/tapedfunction.jl
KDr2: A `Box` or `nothing` 😄

rikhuijzer (commented):
The integration tests of #106 and #109 are not enabled for this PR which can be seen by looking at the running time for the integration tests. Running all tests takes longer than 1 minute.

src/tapedfunction.jl
phipsgabler: Any reason not to use a `Union` then, if only for documentation purposes?

KDr2: Updated.

KDr2:
Oh my, who deleted the master branch? 🤕

yebai:
> Oh my, who deleted the master branch? 🤕

Fixed - it was deleted by `merge-bot` because I merged a PR from `master` to this PR. I also added protection so this won't happen in the future.

src/tapedfunction.jl
```diff
@@ -10,7 +14,25 @@ An `Instruction` stands for a function call
...
+
+mutable struct BlockInstruction{T<:Taped} <: AbstractInstruction
+    id::Int
+    args::Vector
+    tape::T
+end
+
+mutable struct BranchInstruction{T<:Taped} <: AbstractInstruction
+    condition::Union{Bool, Box{Any}, Nothing}
+    block::Int
```
KDr2, line 29:
```suggestion
    block_id::Int
```

src/tapedfunction.jl
```diff
@@ -10,7 +14,25 @@ An `Instruction` stands for a function call
 mutable struct Instruction{F, T<:Taped} <: AbstractInstruction
     func::F
     input::Tuple
-    output
+    output::Box
+    tape::T
+end
+
+mutable struct BlockInstruction{T<:Taped} <: AbstractInstruction
+    id::Int
```
KDr2, line 22:
```suggestion
    block_id::Int
```

src/tapedfunction.jl
```diff
@@ -10,7 +14,25 @@ An `Instruction` stands for a function call
...
+    args::Vector
+    tape::T
+end
+
+mutable struct BranchInstruction{T<:Taped} <: AbstractInstruction
+    condition::Union{Bool, Box{Any}, Nothing}
```
KDr2, line 28: use false for nothing, remove `Nothing`

src/tapedfunction.jl
```diff
@@ -137,122 +176,101 @@ function (instr::Instruction{typeof(_new)})()
...
-        else
-            @warn "Unknown IR code: " st
+        # branches (including `return`)
+        for br in IRTools.branches(blk)
+            if br.condition === nothing && br.block == 0
+                ins = ReturnInstruction(_box(br.args[1]), taped)
+                push!(tape, ins)
+            else
+                ins = BranchInstruction(
+                    _box(br.condition), br.block, map(_box, br.args), taped)
```
KDr2, line 268: use false if br.condition is nothing.

src/tapedfunction.jl
phipsgabler: OK cool. Can you add couple of comments somewhere about that? Or is there some kind of developer documentation elsewhere?

src/tapedfunction.jl
```diff
@@ -42,39 +63,44 @@ Base.show(io::IO, box::Box) = print(io, "Box(", box.val, ")")
...
-result(t::RawTape) = isempty(t) ? nothing : val(t[end].output)
-result(t::TapedFunction) = result(t.tape)
+result(t::TapedFunction) = t.retval
 
-function increase_counter!(t::TapedFunction)
-    t.counter > length(t) && return
-    # instr = t[t.counter]
-    t.counter += 1
-    return t
+function init!(tf::TapedFunction, args)
```
KDr2, line 68: move this into the inner constructor.

src/tapedfunction.jl
```diff
@@ -42,39 +63,44 @@ Base.show(io::IO, box::Box) = print(io, "Box(", box.val, ")")
...
+    if isempty(tf.tape)
+        ir = IRTools.@code_ir tf.func(args...)
+        tf.ir = ir
+        translate!(tf, ir)
+    end
+    return tf
 end
 
 function reset!(tf::TapedFunction, ir::IRTools.IR, tape::RawTape)
```
KDr2, line 77: make this a constructor.

src/tapedtask.jl
```diff
@@ -203,33 +201,56 @@ end
 copy_box(o, roster::Dict{UInt64, Any}) = o
 
 function Base.copy(x::Instruction, on_tape::Taped, roster::Dict{UInt64, Any})
+    func = copy_box(x.func, roster)
```
KDr2, line 204: add a comment

src/tapedfunction.jl
```diff
@@ -7,32 +11,59 @@ const RawTape = Vector{AbstractInstruction}
...
+mutable struct ReturnInstruction{TA, T<:Taped} <: AbstractInstruction
+    arg::TA
     tape::T
 end
 
 mutable struct TapedFunction{F} <: Taped
-    func::F # maybe a function or a callable obejct
+    func::F # maybe a function, a constructor, or a callable obejct
     arity::Int
     ir::Union{Nothing, IRTools.IR}
```
yebai, line 42: I think Julia allows us to leave fields of structs as `#undef`. Maybe we can let `ir` be `#undef` to replace `nothing`. Then, we can replace `Union{Nothing, IRTools.IR}` with `IRTools.IR`.

src/tapedfunction.jl
```diff
@@ -7,32 +11,59 @@ const RawTape = Vector{AbstractInstruction}
...
 end
 
-mutable struct Box{T}
-    val::T
-end
-
 ## methods for Box
 val(x) = x
 val(x::Box) = x.val
+val(x::GlobalRef) = getproperty(x.mod, x.name)
```
yebai, line 65: Nice improvements!

src/tapedfunction.jl
```diff
@@ -138,124 +162,97 @@ function (instr::Instruction{typeof(_new)})()
...
-        for i in 1:arity
-            last_pos = IRTools.insertafter!(ir, last_pos, IRTools.xcall(Base, :indexed_iterate, args_var, i, iter_state...))
-            args_iter = last_pos
-            last_pos = IRTools.insertafter!(ir, last_pos, IRTools.xcall(Core, :getfield, args_iter, 1))
-            push!(args_new, last_pos)
-            if i != arity
-                last_pos = IRTools.insertafter!(ir, last_pos, IRTools.xcall(Core, :getfield, args_iter, 2))
-                iter_state = [last_pos]
-            end
+arg_boxer(var, boxes) = var
```
yebai, line 199:
```suggestion
arg_boxer(var, boxes::Dict) = var
```

src/tapedfunction.jl
```diff
@@ -138,124 +162,97 @@ function (instr::Instruction{typeof(_new)})()
...
+arg_boxer(var::IRTools.Variable, boxes) = get!(boxes, var, Box{Any}(nothing))
```
yebai, line 200:
```suggestion
arg_boxer(var::IRTools.Variable, boxes::Dict) = get!(boxes, var, Box{Any}(nothing))
```

src/tapedfunction.jl
```diff
@@ -138,124 +162,97 @@ function (instr::Instruction{typeof(_new)})()
...
+                push!(tape, ins)
+            elseif Meta.isexpr(st.expr, :new)
+                args = map(_box, st.expr.args)
+                ins = Instruction(_new, args |> Tuple, _box(x), taped)
+                push!(tape, ins)
+            elseif isa(st.expr, GlobalRef)
+                ins = Instruction(val, (st.expr,), _box(x), taped)
+                push!(tape, ins)
+            else
+                @warn "Unknown IR code: " st
```
yebai, line 238: I think we should error here so that user knows something went wrong. 
```suggestion
                @error "Unknown IR code: " st
```

src/tapedtask.jl
```diff
@@ -59,34 +46,44 @@ function TapedTask(tf::TapedFunction, args...)
     return t
 end
 
-# Issue: evaluating model without a trace, see
-# https://github.com/TuringLang/Turing.jl/pull/1757#diff-8d16dd13c316055e55f300cd24294bb2f73f46cbcb5a481f8936ff56939da7ceR329
-TapedTask(f, args...) = TapedTask(TapedFunction(f, arity=length(args)), args...)
+function TapedTask(f, args...)
```
yebai, line 49: Let’s keep this comment for future reference. 
```suggestion
# NOTE: evaluating model without a trace, see
 # https://github.com/TuringLang/Turing.jl/pull/1757#diff-8d16dd13c316055e55f300cd24294bb2f73f46cbcb5a481f8936ff56939da7ceR329
function TapedTask(f, args...)
```

src/tapedtask.jl
```diff
@@ -59,34 +46,44 @@ function TapedTask(tf::TapedFunction, args...)
...
+    cache_key = (f, typeof.(args)...)
+    if haskey(TRCache, cache_key)
```
yebai, line 51: Let’s perhaps move this caching logic into `TapeFunction` constructor. Apologies for the confusion in our meeting. After some thought, I feel this fits better in `tapedfunction.jl`.

src/tapedtask.jl
```diff
@@ -59,34 +46,44 @@ function TapedTask(tf::TapedFunction, args...)
...
+        cached_tf = TRCache[cache_key]
+        # Here we don't need change the initial arguments of the tape,
+        # it will be set when we `step_in` to the tape.
+        tf = copy(cached_tf)
+        tf.counter = 1
+    else
+        tf = TapedFunction(f, args...)
+        TRCache[cache_key] = tf
```
yebai, line 59: We no longer need this if we cache `tapedfunction` by default.

src/tapedtask.jl
```diff
@@ -59,34 +46,44 @@ function TapedTask(tf::TapedFunction, args...)
...
+    end
+    TapedTask(tf, args...)
+end
+
 TapedTask(t::TapedTask, args...) = TapedTask(func(t), args...)
 func(t::TapedTask) = t.tf.func
 
 function step_in(tf::TapedFunction, args)
```
yebai, line 67: I feel we can merge `step_in` with `(tf::FapedFunction)(args)` by adding a callback function, e.g. 

```julia

function (tf::TapedFunction)(args...; callback=nothing)
     # run the raw tape
     if length(args) > 0
         input = map(box, args)
         tf.tape[1].input = input
     end
     tf.counter = 1
     while true
         ins = tf[tf.counter]
         ins()

        callback isa nothing || callback() # we can wrap the produce and wait code into this callback function. 
         isa(ins, ReturnInstruction) && break
     end
     return result(tf)
 end

```

src/tapedtask.jl
```diff
@@ -203,33 +200,57 @@ end
...
+end
+
+function Base.copy(x::ReturnInstruction, on_tape::Taped, roster::Dict{UInt64, Any})
+    arg = copy_box(x.arg, roster)
+    ReturnInstruction(arg, on_tape)
+end
 
-    for (i, x) in enumerate(old_data[start:end])
+function Base.copy(old_data::RawTape, on_tape::Taped, roster::Dict{UInt64, Any})
+    new_data = RawTape(undef, length(old_data))
```
yebai, line 234: Maybe use `new_tape` and `old_tape` for better clarity? 
```suggestion
    new_tape = RawTape(undef, length(old_data))
```

src/tapedtask.jl
```diff
@@ -203,33 +200,57 @@ end
...
         new_data[i] = new_ins
     end
 
+    init_ins = Instruction(args_initializer(new_data[2]), (),
+                           Box{Any}(nothing), on_tape)
+    new_data[1] = init_ins
     return new_data
 end
 
 function Base.copy(tf::TapedFunction)
```
yebai, line 246: Maybe these `copy` functions for `Instruciton`/`TapedFunction`/`RawTape` should probably be moved to the `tapedfunciton.jl` file.

yebai (changes requested):
Many thanks @KDr2 - I took another careful look. See below for comments.

yebai (approved):

PR #110, opened 2022-01-23
Deepcopy `args` in `CTask` constructor to avoid sharing arguments

yebai:
Sharing arguments (e.g. `varinfo`) in `CTask` might be causing some issues if we evaluate the model in parallel on multiple threads (see e.g., [here](https://github.com/TuringLang/Turing.jl/runs/4865925201?check_suite_focus=true#step:6:4409)). This works fine previously because the model evaluator was not shared until https://github.com/TuringLang/Turing.jl/pull/1757. That is, every time model is (re-)evaluated,  `varinfo` is deep-copied implicitly in the `evaluate!!` pipeline (possibly via `DynamicPPL.resetlogp!!(varinfo)`??).  

Update: the[ `reset_model`](https://github.com/TuringLang/AdvancedPS.jl/blob/hg/update-libtask-api/src/container.jl#L32) function seems also handles this while copying particles.

### DIFF
--- a/src/tapedtask.jl
+++ b/src/tapedtask.jl
@@ -61,7 +61,7 @@ end
 
 # Issue: evaluating model without a trace, see
 # https://github.com/TuringLang/Turing.jl/pull/1757#diff-8d16dd13c316055e55f300cd24294bb2f73f46cbcb5a481f8936ff56939da7ceR329
-TapedTask(f, args...) = TapedTask(TapedFunction(f, arity=length(args)), args...)
+TapedTask(f, args...) = TapedTask(TapedFunction(f, arity=length(args)), (deepcopy.(args))...)
 TapedTask(t::TapedTask, args...) = TapedTask(func(t), args...)
 func(t::TapedTask) = t.tf.func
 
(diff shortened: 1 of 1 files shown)


src/tapedtask.jl
```diff
@@ -61,7 +61,7 @@ end
 
 # Issue: evaluating model without a trace, see
 # https://github.com/TuringLang/Turing.jl/pull/1757#diff-8d16dd13c316055e55f300cd24294bb2f73f46cbcb5a481f8936ff56939da7ceR329
-TapedTask(f, args...) = TapedTask(TapedFunction(f, arity=length(args)), args...)
+TapedTask(f, args...) = TapedTask(TapedFunction(f, arity=length(args)), (deepcopy.(args))...)
```
yebai, line 64:
```suggestion
TapedTask(f, args...) = TapedTask(TapedFunction(f, arity=length(args)), deepcopy.(args)...)
```

yebai:
This requires a bit more discussion - please don't merge yet.

KDr2:
I think it's the caller's responsibility to make sure they pass the correct data/arguments. Deeply copying the arguments in all cases may not be what the caller wants.

yebai:
That sounds sensible!

PR #114, opened 2022-01-30
Deepcopy array and dict by default when copying tape

yebai:


yebai:
Closed by https://github.com/TuringLang/Libtask.jl/pull/116

PR #118, opened 2022-02-15
Fix non-concrete type near TRCache

rikhuijzer:
Very minor change. This fixes non-concrete types reported by
```julia
TapedTask(f, args...)
```
and
```julia
TapedFunction(f, args...; cache=true, init=true)
```
The compiler wasn't able to determine the type of `TRCache[cache_key]` which makes sense.

Also, I've tried to narrow down the types of `TRCache` which makes things slightly easier for the readers and the compiler.

Same for `TapedTaskException.backtrace`.

### DIFF
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -36,10 +36,8 @@ mutable struct ReturnInstruction{TA, T<:Taped} <: AbstractInstruction
     tape::T
 end
 
-const TRCache = LRU{Any, Any}(maxsize=10)
-
 mutable struct TapedFunction{F} <: Taped
-    func::F # maybe a function, a constructor, or a callable obejct
+    func::F # maybe a function, a constructor, or a callable object
     arity::Int
     ir::IRTools.IR
     tape::RawTape
@@ -53,7 +51,7 @@ mutable struct TapedFunction{F} <: Taped
         cache_key = (f, typeof.(args)...)
 
         if cache && haskey(TRCache, cache_key) # use cache
-            cached_tf = TRCache[cache_key]
+            cached_tf = TRCache[cache_key]::TapedFunction{F}
             tf = copy(cached_tf)
             tf.counter = 1
             return tf
@@ -76,6 +74,8 @@ mutable struct TapedFunction{F} <: Taped
         return tf
     end
 end
+
+const TRCache = LRU{Tuple, TapedFunction}(maxsize=10)
 
 ## methods for Box
 val(x) = x
--- a/src/tapedtask.jl
+++ b/src/tapedtask.jl
@@ -1,6 +1,6 @@
 struct TapedTaskException
     exc::Exception
-    backtrace
+    backtrace::Vector{Any}
 end
 
 struct TapedTask


devmotion:
The types, e.g., of the cache, are still non-concrete though. Somewhere I read recently that there are cases where it is better to make/leave non-concrete types as `Any` instead of using another more restrictive non-concrete type... I am not sure if it applies to any of the changes in this PR though.

It sounds like this PR is supposed to fix some type inference issue though? Can we add tests for it?

src/tapedfunction.jl
```diff
@@ -53,7 +51,7 @@ mutable struct TapedFunction{F} <: Taped
         cache_key = (f, typeof.(args)...)
 
         if cache && haskey(TRCache, cache_key) # use cache
-            cached_tf = TRCache[cache_key]
+            cached_tf = TRCache[cache_key]::TapedFunction{F}
```
devmotion, line 54: I don't know if one could come up with an example that breaks this assertion. But it seems reasonable.

rikhuijzer:
> It sounds like this PR is supposed to fix some type inference issue though? Can we add tests for it?

I think yes. Test via `JETTest` or is there some other method?

I’ll respond to your other comments tomorrow

yebai (approved):
thanks @rikhuijzer - it looks good to me!

yebai:
It'll be great to gradually fix the type instability issues in `TapedFunction`. It also helps to keep these PRs small (this PR is an excellent example) so they can be reviewed and merged quickly.

rikhuijzer:
> The types, e.g., of the cache, are still non-concrete though. Somewhere I read recently that there are cases where it is better to make/leave non-concrete types as `Any` instead of using another more restrictive non-concrete type... I am not sure if it applies to any of the changes in this PR though.

Maybe, you've read https://github.com/SciML/ModelingToolkit.jl/pull/1215#issuecomment-902258784? I've tried creating the `TRCache` via
```julia
const TRCache = LRU{Tuple, TapedFunction{Any}}(maxsize=10)
```
but that gave a "could not convert" errror.

devmotion:
> but that gave a "could not convert" errror.

That's expected since `TapedFunction{F}` for a concrete `F` (and even `TapedFunction{F} where F`) is not a subtype of `TapedFunction{Any}`, so it would have to be converted: https://docs.julialang.org/en/v1/manual/types/#man-parametric-composite-types

> Maybe, you've read [SciML/ModelingToolkit.jl#1215 (comment)](https://github.com/SciML/ModelingToolkit.jl/pull/1215#issuecomment-902258784)?

Hmm, I don't think so, but it was a similar comment somewhere. In any case, with respect to the issue mentioned in the PR the change in this PR is no improvement, since the cache already satisfied `isconcretetype`:
```julia
julia> using LRUCache

julia> cache1 = LRU{Any,Any}(maxsize=10);

julia> typeof(cache1)
LRU{Any, Any}

julia> isconcretetype(typeof(cache1))
true

julia> struct T{F}
           f::F
       end

julia> cache2 = LRU{Tuple,T}(maxsize=10);

julia> isconcretetype(typeof(cache2))
true

julia> typeof(cache2)
LRU{Tuple, T}
```

Of course, the PR changes the (non-)inferred return type of `cache[key]` (in both cases it can't be inferred exactly):
```julia
julia> cache1[(:a, 1)] = T(2.0)
T{Float64}(2.0)

julia> @code_warntype cache1[(:a, 1)]
MethodInstance for getindex(::LRU{Any, Any}, ::Tuple{Symbol, Int64})
  from getindex(lru::LRU, key) in LRUCache at /home/david/.julia/packages/LRUCache/44dJX/src/LRUCache.jl:125
Arguments
  #self#::Core.Const(getindex)
  lru::LRU{Any, Any}
  key::Tuple{Symbol, Int64}
Locals
  #12::LRUCache.var"#12#13"{LRU{Any, Any}, Tuple{Symbol, Int64}}
Body::Any
1 ─ %1 = LRUCache.:(var"#12#13")::Core.Const(LRUCache.var"#12#13")
│   %2 = Core.typeof(lru)::Core.Const(LRU{Any, Any})
│   %3 = Core.typeof(key)::Core.Const(Tuple{Symbol, Int64})
│   %4 = Core.apply_type(%1, %2, %3)::Core.Const(LRUCache.var"#12#13"{LRU{Any, Any}, Tuple{Symbol, Int64}})
│        (#12 = %new(%4, lru, key))
│   %6 = #12::LRUCache.var"#12#13"{LRU{Any, Any}, Tuple{Symbol, Int64}}
│   %7 = Base.getproperty(lru, :lock)::Base.Threads.SpinLock
│   %8 = LRUCache.lock(%6, %7)::Any
└──      return %8

julia> cache2[(:a, 1)] = T(2.0)
T{Float64}(2.0)

julia> @code_warntype cache2[(:a, 1)]
MethodInstance for getindex(::LRU{Tuple, T}, ::Tuple{Symbol, Int64})
  from getindex(lru::LRU, key) in LRUCache at /home/david/.julia/packages/LRUCache/44dJX/src/LRUCache.jl:125
Arguments
  #self#::Core.Const(getindex)
  lru::LRU{Tuple, T}
  key::Tuple{Symbol, Int64}
Locals
  #12::LRUCache.var"#12#13"{LRU{Tuple, T}, Tuple{Symbol, Int64}}
Body::T
1 ─ %1 = LRUCache.:(var"#12#13")::Core.Const(LRUCache.var"#12#13")
│   %2 = Core.typeof(lru)::Core.Const(LRU{Tuple, T})
│   %3 = Core.typeof(key)::Core.Const(Tuple{Symbol, Int64})
│   %4 = Core.apply_type(%1, %2, %3)::Core.Const(LRUCache.var"#12#13"{LRU{Tuple, T}, Tuple{Symbol, Int64}})
│        (#12 = %new(%4, lru, key))
│   %6 = #12::LRUCache.var"#12#13"{LRU{Tuple, T}, Tuple{Symbol, Int64}}
│   %7 = Base.getproperty(lru, :lock)::Base.Threads.SpinLock
│   %8 = LRUCache.lock(%6, %7)::T
└──      return %8
```

rikhuijzer:
> Hmm, I don't think so, but it was a similar comment somewhere. In any case, with respect to the issue mentioned in the PR the change in this PR is no improvement, since the cache already satisfied `isconcretetype`:

Agreed, but it is an improvement with regards to readability.

> Of course, the PR changes the (non-)inferred return type of `cache[key]` (in both cases it can't be inferred exactly):

Yes also correct.

Let's merge?

devmotion (approved):

devmotion:
Maybe bump the version number in the PR, otherwise looks good to me :slightly_smiling_face:

rikhuijzer:
> Maybe bump the version number in the PR, otherwise looks good to me :slightly_smiling_face:

Oh yes, I forgot that that is TuringLang convention. Done now.

Thank you for the review! I don't have merge rights here, can one of you do it please?

yebai:
Thanks, @rikhuijzer - you should have merge rights now.

PR #120, opened 2022-02-18
Extract TapedTask try-catch method

rikhuijzer:
I'm still working a bit through the code in order to find performance gains. This PR applies an extract method refactoring to make the `TapedTask` easier to read. It has no negative nor positive impact on performance. I've verified this by putting a `@show` on the `TapedTask(tf, args...)` call and that shows `(32 allocations: 2.312 KiB)` in both versions. 

Because the git diff is very unhelpful, here are the before and after:

## master branch 

```julia
function TapedTask(tf::TapedFunction, args...)
    produce_ch = Channel()
    consume_ch = Channel{Int}()
    task = @task try
        producer = () -> begin
            ttask = current_task().storage[:tapedtask]
            if length(ttask.produced_val) > 0
                val = pop!(ttask.produced_val)
                put!(ttask.produce_ch, val)
                take!(ttask.consume_ch) # wait for next consumer
            end
        end
        tf(args...; callback=producer)
    catch e
        bt = catch_backtrace()
        put!(produce_ch, TapedTaskException(e, bt))
        # @error "TapedTask Error: " exception=(e, bt)
        rethrow()
    finally
        @static if VERSION >= v"1.4"
            # we don't do this under Julia 1.3, because `isempty` always hangs on
            # an empty channel.
            while !isempty(produce_ch)
                yield()
            end
        end
        close(produce_ch)
        close(consume_ch)
    end
    t = TapedTask(task, tf, produce_ch, consume_ch)
    task.storage === nothing && (task.storage = IdDict())
    task.storage[:tapedtask] = t
    tf.owner = t
    return t
end
``` 

## this PR

```julia
function wrap_task(tf, produce_ch, consume_ch, args...)
    try
        producer = () -> begin
            ttask = current_task().storage[:tapedtask]
            if length(ttask.produced_val) > 0
                val = pop!(ttask.produced_val)
                put!(ttask.produce_ch, val)
                take!(ttask.consume_ch) # wait for next consumer
            end
        end
        tf(args...; callback=producer)
    catch e
        bt = catch_backtrace()
        put!(produce_ch, TapedTaskException(e, bt))
        # @error "TapedTask Error: " exception=(e, bt)
        rethrow()
    finally
        @static if VERSION >= v"1.4"
            # we don't do this under Julia 1.3, because `isempty` always hangs on
            # an empty channel.
            while !isempty(produce_ch)
                yield()
            end
        end
        close(produce_ch)
        close(consume_ch)
    end
end

function TapedTask(tf::TapedFunction, args...)
    produce_ch = Channel()
    consume_ch = Channel{Int}()
    task = @task wrap_task(tf, produce_ch, consume_ch, args...)
    t = TapedTask(task, tf, produce_ch, consume_ch)
    task.storage === nothing && (task.storage = IdDict())
    task.storage[:tapedtask] = t
    tf.owner = t
    return t
end
```

### DIFF
--- a/src/tapedtask.jl
+++ b/src/tapedtask.jl
@@ -16,10 +16,8 @@ struct TapedTask
     end
 end
 
-function TapedTask(tf::TapedFunction, args...)
-    produce_ch = Channel()
-    consume_ch = Channel{Int}()
-    task = @task try
+function wrap_task(tf, produce_ch, consume_ch, args...)
+    try
         producer = () -> begin
             ttask = current_task().storage[:tapedtask]
             if length(ttask.produced_val) > 0
@@ -45,6 +43,12 @@ function TapedTask(tf::TapedFunction, args...)
         close(produce_ch)
         close(consume_ch)
     end
+end
+
+function TapedTask(tf::TapedFunction, args...)
+    produce_ch = Channel()
+    consume_ch = Channel{Int}()
+    task = @task wrap_task(tf, produce_ch, consume_ch, args...)
     t = TapedTask(task, tf, produce_ch, consume_ch)
     task.storage === nothing && (task.storage = IdDict())
     task.storage[:tapedtask] = t


yebai (approved):

yebai:
It makes sense to me to separate the task function for better clarity.

PR #125, opened 2022-02-20
Remove redundant new function.

yebai:


yebai:
Test error is irrelevant to this PR.

rikhuijzer (commented):
Actually, `_new` and `__new__` are two different things. 

`_new` is a very weird (in my opinion) function which denotes that an instruction is new. It is used to dispatch on `(instr)()` where `instr::Instruction{typeof(_new})`. This works because `Instruction` has that parametric type which can hold anything, so also some function. Given the setting, I get it but it took me a while to figure out what it is.

`__new__` on the other hand, is a constructor for creating objects.

So, maybe we should rename `__new__` to `create_object` or something? To avoid confusion

yebai:
Sorry for the confusion. I encouraged @KDr2 to implement a generated function when we added `_new`, but we didn't find the trick that you later added. So `_new` ended up being a type for dispatching purposes only. That is why I combined it with `__new__` in this PR. 

Personally, I find `__new__` more readable than `create_object`. But it is a very minor naming issue so I'm happy with whatever changes we ended up with.

rikhuijzer (approved):
Okay :+1: Then it makes sense and I agree that `__new__` is better in that case.

Issue #117, opened 2022-02-09
Replace `IRTools` with `CodeInfoTools` 


yebai:

devmotion:
Is it better? I am not familiar with the differences (according to the README it is based closely on IRTools) but it seems it is developed less actively and has fewer contributors (and is only used by Mixtape.jl but no other package yet).

yebai:
`CodeInfoTool` is based on the new compiler data structures and APIs in Julia 1.7. Since `CodeInfoTools` is just a small set of utilities for making Julia's compiler API more friendly, it should be more stable than `IRTools` in general - that's probably why `CodeInfoTools` does not need any changes for a long time. In the worst case, it should not be hard to switch using Julia's native compiler API from `CodeInfoTools`. However, it is slightly less user-friendly.

femtomc:
> `CodeInfoTool` is based on the new compiler data structures (i.e. `CodeInfo`) and APIs in Julia 1.7. Since `CodeInfoTools` is just a small set of utilities for making Julia's compiler API more friendly, it should be more stable than `IRTools` in general - that's probably why `CodeInfoTools` does not need any changes for a long time. In the worst case, it should not be hard to switch to using Julia's native compiler API from `CodeInfoTools`. However, Julia's compiler API is slightly less user-friendly in its current form.

This is the right idea. CodeInfoTools should be thought of as an intermediate step before better APIs are actually part of Core.

PR #126, opened 2022-02-23
Switch from IRTools to CodeInfo

KDr2:
Fix https://github.com/TuringLang/Libtask.jl/issues/117

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,19 +5,17 @@ desc = "Tape based task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
 version = "0.7.0"
 
-
 [deps]
-IRTools = "7869d1d1-7146-5819-86e3-90919afe41df"
+CodeInfoTools = "bc773b8a-8374-437a-b9f2-0e9785855863"
 LRUCache = "8ac3fa9e-de4c-5943-b1dc-09c6b5f20637"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-julia = "1.3"
-MacroTools = "0.5"
-IRTools = "0.4"
 LRUCache = "1.3"
+MacroTools = "0.5"
+julia = "1.3"
 
 [extras]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
--- a/Project.toml
+++ b/Project.toml
@@ -13,6 +13,7 @@ MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
+CodeInfoTools = "0.3.4"
 LRUCache = "1.3"
 MacroTools = "0.5"
 julia = "1.7"
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 license = "MIT"
 desc = "Tape based task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
-version = "0.8.0"
+version = "0.7.0"
 
 [deps]
 CodeInfoTools = "bc773b8a-8374-437a-b9f2-0e9785855863"
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 license = "MIT"
 desc = "Tape based task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
-version = "0.7.0"
+version = "0.8.0"
 
 [deps]
 CodeInfoTools = "bc773b8a-8374-437a-b9f2-0e9785855863"
@@ -15,7 +15,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [compat]
 LRUCache = "1.3"
 MacroTools = "0.5"
-julia = "1.6"
+julia = "1.7"
 
 [extras]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
--- a/Project.toml
+++ b/Project.toml
@@ -15,7 +15,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [compat]
 LRUCache = "1.3"
 MacroTools = "0.5"
-julia = "1.3"
+julia = "1.6"
 
 [extras]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -186,8 +186,7 @@ end
 
 function (instr::GotoInstruction)()
     if instr.condition === nothing || !val(instr.condition) # unless
-        target = instr.dest
-        instr.tape.counter = target
+        instr.tape.counter = instr.dest
     else
         instr.tape.counter += 1
     end
@@ -258,12 +257,12 @@ function translate!(taped::Taped, ir::Core.CodeInfo)
         elseif isa(line, Core.ReturnNode)
             ins = ReturnInstruction(_box(line.val), taped)
             push!(tape, ins)
+        elseif isa(line, GlobalRef)
+            ins = Instruction(val, (line,), _box(Core.SSAValue(idx)), taped)
+            push!(tape, ins)
         elseif Meta.isexpr(line, :new)
             args = map(_box, line.args)
             ins = Instruction(__new__, args |> Tuple, _box(Core.SSAValue(idx)), taped)
-            push!(tape, ins)
-        elseif isa(line, GlobalRef)
-            ins = Instruction(val, (line,), _box(Core.SSAValue(idx)), taped)
             push!(tape, ins)
         elseif Meta.isexpr(line, :(=))
             output = _box(line.args[1]) # args[1] is a SlotNumber
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -1,4 +1,5 @@
 mutable struct Box{T}
+    id::Symbol
     val::T
 end
 
@@ -18,16 +19,9 @@ mutable struct Instruction{F, TI<:Tuple, TO, T<:Taped} <: AbstractInstruction
     tape::T
 end
 
-mutable struct BlockInstruction{TA, T<:Taped} <: AbstractInstruction
-    block_id::Int
-    args::Vector{TA}
-    tape::T
-end
-
-mutable struct BranchInstruction{TA, T<:Taped} <: AbstractInstruction
+mutable struct GotoInstruction{T<:Taped} <: AbstractInstruction
     condition::Box{Any}
-    block_id::Int
-    args::Vector{TA}
+    dest::Int
     tape::T
 end
 
@@ -39,11 +33,9 @@ end
 mutable struct TapedFunction{F} <: Taped
     func::F # maybe a function, a constructor, or a callable object
     arity::Int
-    ir::IRTools.IR
+    ir::Core.CodeInfo
     tape::RawTape
     counter::Int
-    # map from BlockInstruction.block_id to its index on tape
-    block_map::Dict{Int, Int}
     retval
     owner
 
@@ -61,11 +53,10 @@ mutable struct TapedFunction{F} <: Taped
         tf.func = f
         tf.tape = RawTape()
         tf.counter = 1
-        tf.block_map = Dict{Int, Int}()
 
         if init # init
             tf.arity = length(args)
-            ir = IRTools.@code_ir tf.func(args...)
+            ir = code_inferred(tf.func, typeof.(args)...)
             tf.ir = ir
             translate!(tf, ir)
             # set cache
@@ -83,9 +74,12 @@ val(x::Box) = x.val
 val(x::GlobalRef) = getproperty(x.mod, x.name)
 val(x::QuoteNode) = eval(x)
 val(x::TapedFunction) = x.func
-box(x) = Box(x)
 box(x::Box) = x
-Base.show(io::IO, box::Box) = print(io, "Box(", box.val, ")")
+box(x) = Box(gensym(), x)
+box(id::Symbol, x) = Box(id, x)
+box_any(x) = Box{Any}(gensym(), x)
+box_any(id::Symbol, x) = Box{Any}(id, x)
+Base.show(io::IO, box::Box) = print(io, "Box(", box.id, ", ", box.val, ")")
 
 ## methods for RawTape and Taped
 MacroTools.@forward TapedFunction.tape Base.iterate, Base.length
@@ -96,7 +90,7 @@ result(t::TapedFunction) = t.retval
 function (tf::TapedFunction)(args...; callback=nothing)
     # run the raw tape
     if(tf.counter <= 1 && length(args) > 0)
-        input = map(Box{Any}, args)
+        input = map(box_any, args)
         tf.tape[1].input = input
     end
 
@@ -149,15 +143,9 @@ function Base.show(io::IO, instruction::Instruction)
     println(io, "Instruction($(func)$(map(val, instruction.input)), tape=$(objectid(tape)))")
 end
 
-function Base.show(io::IO, instruction::BlockInstruction)
-    id = instruction.block_id
+function Base.show(io::IO, instruction::GotoInstruction)
     tape = instruction.tape
-    println(io, "BlockInstruction($(id)->$(map(val, instruction.args)), tape=$(objectid(tape)))")
-end
-
-function Base.show(io::IO, instruction::BranchInstruction)
-    tape = instruction.tape
-    println(io, "BranchInstruction($(val(instruction.condition)), tape=$(objectid(tape)))")
+    println(io, "GotoInstruction($(val(instruction.condition)), tape=$(objectid(tape)))")
 end
 
 function (instr::Instruction{F})() where F
@@ -199,21 +187,10 @@ function (instr::Instruction{typeof(__new__)})()
     end
 end
 
-function (instr::BlockInstruction)()
-    instr.tape.counter += 1
-end
-
-function (instr::BranchInstruction)()
+function (instr::GotoInstruction)()
     if instr.condition === nothing || !val(instr.condition) # unless
-        target = instr.block_id
-        target_idx = instr.tape.block_map[target]
-        blk_ins = instr.tape[target_idx]
-        @assert isa(blk_ins, BlockInstruction)
-        @assert length(instr.args) == length(blk_ins.args)
-        for i in 1:length(instr.args)
-            blk_ins.args[i].val = val(instr.args[i])
-        end
-        instr.tape.counter = target_idx
+        target = instr.dest
+        instr.tape.counter = target
     else
         instr.tape.counter += 1
     end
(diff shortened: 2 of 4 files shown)


KDr2:
It turns out `CodeInfo` differs **REALLY A LOT** between versions...

yebai:
> It turns out CodeInfo differs REALLY A LOT between versions...

Let's focus on Julia `>=1.7` in this PR.

KDr2:
Take function `g` below as an exmaple:
```
function g(x, y)
  if x>y
    r= string(sin(x))
  else
    r= sin(x) * cos(y)
  end
  return r
end
```

Its `CodeInfo` under 1.6:
```
 :(Base.slt_int(_3, _2))
 :(goto %7 if not %1)
 :(Base.sitofp(Float64, _2))
 :($(Expr(:invoke, MethodInstance for sin(::Float64), :(Base.Math.sin), :(%3))))
 :($(Expr(:invoke, MethodInstance for string(::Float64), :(Main.string), :(%4))))
 :(goto %12)
 :(Base.sitofp(Float64, _2))
 :($(Expr(:invoke, MethodInstance for sin(::Float64), :(Base.Math.sin), :(%7))))
 :(Base.sitofp(Float64, _3))
 :($(Expr(:invoke, MethodInstance for cos(::Float64), :(Base.Math.cos), :(%9))))
 :(Base.mul_float(%8, %10))
 :(φ (%6 => %5, %11 => %11))
 :(return %12)
```

Under 1.7
```
 Core.NewvarNode(:(_4))
 :(_2 > _3)
 :(goto %7 if not %2)
 :(Main.sin(_2))
 :(_4 = Main.string(%4))
 :(goto %10)
 :(Main.sin(_2))
 :(Main.cos(_3))
 :(_4 = %7 * %8)
 :(return _4)
```

Under 1.8:
`CodeInfoToosl.jl` throws en error while getting its `CodeInfo`, code fetched by `@code_lwoered`:

```
julia> @code_lowered g(1.,2.)
CodeInfo(
1 ─      Core.NewvarNode(:(r))
│   %2 = x > y
└──      goto #3 if not %2
2 ─ %4 = Main.sin(x)
│        r = Main.string(%4)
└──      goto #4
3 ─ %7 = Main.sin(x)
│   %8 = Main.cos(y)
└──      r = %7 * %8
4 ┄      return r
)
```

Conclusion: 
- `CodeInfo` changed a lot between v1.6 and v1.7
- `CodeInfoTools.jl` doesn't work on v1.8

yebai:
> CodeInfo changed a lot between v1.6 and v1.7

Good to know,  probably due to the major compiler refactoring in 1.7.

> CodeInfoTools.jl doesn't work on v1.8

Maybe raise an issue on the `CodeInfoTools` package, or help fix it if not major?

KDr2:
Maybe we can remove the dependency of `CodeInfoTools.jl` too, all we want is to get the `CodeInfo` yet, but not to update the `CodeInfo`, so maybe there's a simple way to do that.

yebai:
> Maybe we can remove the dependency of CodeInfoTools.jl too, all we want is to get the CodeInfo yet, but not to update the CodeInfo, so maybe there's a simple way to do that.

We will likely need to update `CodeInfo` in the future, e.g. when we inline all the instructions. So we will probably build up a set of utilities for manipulating `CodeInfo`, which will be similar to `CodeInfoTools`. If that's the case, it might be better to depend on `CodeInfoTools` and help to improve it. 

I took a quick look at the 1.8 erro, and itt appears that Julia has a simple function signature change between 1.7 and 1.8. It should be possible to fix that quickly.

yebai:
CC @femtomc

Project.toml
```diff
@@ -5,19 +5,17 @@ desc = "Tape based task copying in Turing"
...
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
-julia = "1.3"
-MacroTools = "0.5"
-IRTools = "0.4"
 LRUCache = "1.3"
+MacroTools = "0.5"
```
devmotion, line 17: It was an indirect dependency of IRTools anyway but now that it is removed - could MacroTools be removed?

Project.toml
```diff
@@ -5,19 +5,17 @@ desc = "Tape based task copying in Turing"
...
+julia = "1.6"
```
devmotion, line 18: The automerge guidelines demand that the version is bumped to 0.8.0 if the Julia compat is narrowed: Patch releases are not allowed to do that (https://juliaregistries.github.io/RegistryCI.jl/stable/guidelines/#New-versions-of-existing-packages).

src/tapedfunction.jl
```diff
@@ -61,11 +53,10 @@ mutable struct TapedFunction{F} <: Taped
         tf.func = f
         tf.tape = RawTape()
         tf.counter = 1
-        tf.block_map = Dict{Int, Int}()
 
         if init # init
             tf.arity = length(args)
-            ir = IRTools.@code_ir tf.func(args...)
+            ir = code_inferred(tf.func, typeof.(args)...)
```
devmotion, line 59: Simpler and without potentially slow broadcasting and splatting:
```suggestion
            ir = code_inferred(tf.func, typeof(args))
```
(One can provide the type of the arguments also as a `Tuple{T1,T2,...}`).

src/tapedfunction.jl
```diff
@@ -83,9 +74,12 @@ val(x::Box) = x.val
 val(x::GlobalRef) = getproperty(x.mod, x.name)
 val(x::QuoteNode) = eval(x)
 val(x::TapedFunction) = x.func
-box(x) = Box(x)
 box(x::Box) = x
-Base.show(io::IO, box::Box) = print(io, "Box(", box.val, ")")
+box(x) = Box(gensym(), x)
+box(id::Symbol, x) = Box(id, x)
+box_any(x) = Box{Any}(gensym(), x)
+box_any(id::Symbol, x) = Box{Any}(id, x)
```
devmotion, line 81: Couldn't you just define a single argument outer constructor?
```suggestion
Box(x) = Box(gensym(), x)
Box{T}(x) where {T} = Box{T}(gensym(), x)
```
?

src/tapedfunction.jl
```diff
@@ -226,70 +203,106 @@ end
...
+    box_id = Symbol("_$(slot)")
+    for ins in tape
+        isa(ins, Instruction) || continue
+        for b in ins.input
+            isa(b, Box) && b.id === box_id && return b
+        end
+        ins.output.id === box_id && return ins.output
+    end
+    slot == 1 && return box_any(nothing) # func is not used
+    @error "None box found for slot" slot
```
devmotion, line 223: Maybe this should be an actual `error`?

femtomc:
Re — open issue in CodeInfoTools with stacktrace for failure on > 1.8.

KDr2:
> Re — open issue in CodeInfoTools with stacktrace for failure on > 1.8.

I opened 2 related isssus:
- https://github.com/JuliaCompilerPlugins/CodeInfoTools.jl/issues/13
- https://github.com/JuliaCompilerPlugins/CodeInfoTools.jl/issues/12

Project.toml
KDr2: `MacroTools.@forward` is still in use.

devmotion: If this is the only place where it is used it will be easy to remove it in a subsequent PR :slightly_smiling_face:

KDr2: I created an issue: https://github.com/TuringLang/Libtask.jl/issues/127

KDr2:
All tests passed with https://github.com/JuliaCompilerPlugins/CodeInfoTools.jl/pull/14.

femtomc:
Re — inlining transforms should likely not live in CodeInfoTools. Ideally, the plugin (speculative) interface should be used to do this. It’s still WIP tho.

also, tracking types and things that are required to do inlining is possible in CodeInfoTools (just like IRTools) but there’s no really code to do that yet

rikhuijzer:
The tests currently fail on the Turing test suite with the following errors (maybe more)

```
┌ Error: Unknown IR code:
│   typeof(line) = Core.SlotNumber
│   line = :(_28)
└ @ Libtask ~/git/Libtask.jl/src/tapedfunction.jl:291
```
and
```
dynamic model: Error During Test at /home/rik/git/Turing.jl/test/test_utils/staging.jl:42
  Got exception outside of a @test
  MethodError: no method matching iterate(::Nothing)
  Closest candidates are:
    iterate(::Union{LinRange, StepRangeLen}) at /nix/store/y46fn31jpgwkzj6hi1xjv0246x5w5dig-julia-bin-1.7.2/share/julia/base/range.jl:826
    iterate(::Union{LinRange, StepRangeLen}, ::Integer) at /nix/store/y46fn31jpgwkzj6hi1xjv0246x5w5dig-julia-bin-1.7.2/share/julia/base/range.jl:826
    iterate(::T) where T<:Union{Base.KeySet{<:Any, <:Dict}, Base.ValueIterator{<:Dict}} at /nix/store/y46fn31jpgwkzj6hi1xjv0246x5w5dig-julia-bin-1.7.2/share/julia/base/dict.jl:695
    ...
  Stacktrace:
    [1] consume(ttask::TapedTask{var"#imm#31"})
      @ Libtask ~/git/Libtask.jl/src/tapedtask.jl:144
```

KDr2:
> The tests currently fail on the Turing test suite with the following errors (maybe more)

Hope https://github.com/TuringLang/Libtask.jl/pull/126/commits/2e9694087b61865bd745855e9f6cb5f84d28738a fixes them.

yebai:
Many thanks, @femtomc would you like to make a release of `CodeInfoTools` so that tests can pass here?

devmotion:
> PS: Why our IntegrationTest doesn't show the errors?

Because Turing is not compatible with Libtask 0.7 or 0.8, and hence it is not possible to run them:
https://github.com/TuringLang/Libtask.jl/runs/5321568590?check_suite_focus=true#step:6:18

rikhuijzer:
> Because Turing is not compatible with Libtask 0.7 or 0.8, and hence it is not possible to run them:
> https://github.com/TuringLang/Libtask.jl/runs/5321568590?check_suite_focus=true#step:6:18

We could add a bit of logic to the CI which removes the compat entry from Turing for these integration tests? Would require only a few lines of code. Ping me if I need to open a PR for this on this repo.

yebai:
> We could add a bit of logic to the CI which removes the compat entry from Turing for these integration tests? Would require only a few lines of code. Ping me if I need to open a PR for this on this repo.

Seconded - that would be nice! @rikhuijzer

devmotion:
> We could add a bit of logic to the CI which removes the compat entry from Turing for these integration tests? Would require only a few lines of code. Ping me if I need to open a PR for this on this repo.

I think we shouldn't do this. The point of breaking releases is to mark that there are breaking changes - so one can't expect downstream packages to work. And if downstream packages are broken it's not an issue since they explicitly state that they might not be compatible with this breaking release and also the package manager will never install them before the compat bound is updated.

devmotion:
And if you are about to make a breaking release you can already now only bump the version in the final commit before the PR is merged - then the downstream packages will be tested and you will see test errors. In this PR the issue is that Turing is not even compatible with Libtask 0.7 though.

Project.toml
```diff
@@ -3,21 +3,19 @@ uuid = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 license = "MIT"
 desc = "Tape based task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
-version = "0.7.0"
-
+version = "0.8.0"
```
yebai, line 6: Let's change this version later so the integration tests can run. 
```suggestion
version = "0.7.0"
```

src/tapedfunction.jl
```diff
@@ -226,70 +202,113 @@ end
...
-            if Meta.isexpr(st.expr, :call)
-                args = map(_box, st.expr.args)
-                # args[1] is the function
+    goto_offset = 1
+    mi = ir.parent
+
+    for (idx, line) in enumerate(ir.code)
+        isa(line, Core.Const) && (line = line.val) # unbox Core.Const
+
+        if isa(line, Core.NewvarNode)
```
rikhuijzer, line 245: Shouldn't this be dispatches on type?

src/tapedfunction.jl
```diff
@@ -226,70 +202,113 @@ end
...
-            else
-                cond = br.condition === nothing ? false : _box(br.condition)
-                isa(cond, Bool) && (cond = Box{Any}(cond)) # unify the condiftion type
-                ins = BranchInstruction(
-                    cond, br.block, map(_box, br.args), taped)
-                push!(tape, ins)
+        elseif Meta.isexpr(line, :call)
+            args = map(_box, line.args)
+            # args[1] is the function
+            f = args[1]
```
rikhuijzer, line 294:
```suggestion
            func = args[1]
```

src/tapedfunction.jl
KDr2: Maybe, it will be many arguments for the would-be separated functions. This if/else branches-hell is clearer to me than separated functions while I was writing it. But for reading, separated functions may be clearer?

Let's see what other guys will say :)

rikhuijzer: > But for reading, separated functions may be clearer?

Also for performance and extensibility. I understand that it's easier while writing and I'm okay with rewriting it to multiple functions in a separate PR

KDr2:
I find that as `__new__` is now a generated function, we don't need to specialize the execution of `Instruction{typeof(__new__)}` any more.

This can be back ported to 0.7.x.

KDr2:
Finally, all `Turing.jl` tests passed, let's wait for the CI task. 🎉

Project.toml
```diff
@@ -5,19 +5,17 @@ desc = "Tape based task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
 version = "0.7.0"
 
-
 [deps]
-IRTools = "7869d1d1-7146-5819-86e3-90919afe41df"
+CodeInfoTools = "bc773b8a-8374-437a-b9f2-0e9785855863"
```
devmotion, line 9: Can you add a compat entry for it?

src/tapedfunction.jl
KDr2: I am doing this now, will commit it by tomorrow's end.

Project.toml
KDr2: Sure, is this correct?
```diff
 [compat]
+CodeInfoTools = "0.3.4"
 LRUCache = "1.3"
 MacroTools = "0.5"
 julia = "1.7"

```

rikhuijzer (approved):
There are some minor details regarding the compat entry still and it would be good to do some benchmarks with the current implementation but to me this PR looks like an improvement, so I'll approve.

Would be good to wait on a double check from Hong and David

Project.toml
devmotion: I didn't keep track of which version is sufficient but it seems tests pass with 0.3.4 so it looks good.

src/tapedfunction.jl
KDr2: It's easier than I thought, I just added a new commit to resolve this.

KDr2:
> Would be good to wait on a double check from Hong and David

Hope I can have a good sleep tonight 🤣

src/tapedfunction.jl
```diff
@@ -168,127 +154,180 @@ function (instr::Instruction{F})() where F
...
-    end
-end
+arg_boxer(var, boxes::Dict{Symbol, Box{<:Any}}) = var
+arg_boxer(var::Core.SSAValue, boxes::Dict{Symbol, Box{<:Any}}) = arg_boxer(Symbol(var.id), boxes)
+arg_boxer(var::Core.TypedSlot, boxes::Dict{Symbol, Box{<:Any}}) =
+    arg_boxer(Symbol("_$(var.id)"), boxes)
+arg_boxer(var::Core.SlotNumber, boxes::Dict{Symbol, Box{<:Any}}) =
+    arg_boxer(Symbol("_$(var.id)"), boxes)
+arg_boxer(var::Symbol, boxes::Dict{Symbol, Box{<:Any}}) =
+    get!(boxes, var, Box{Any}(var, nothing))
```
devmotion, line 202: Not sure how much it helps but one can avoid constructing the default value in every call:

```suggestion
    get!(boxes, var) do
        return Box{Any}(var, nothing)
    end
```

src/tapedfunction.jl
```diff
@@ -168,127 +154,180 @@ function (instr::Instruction{F})() where F
...
         for i in 1:length(args)
-            ins.args[i + 1].val = val(args[i]) # fill the boxes
+            argsbox[i].val = val(args[i]) # fill the boxes
         end
     end
 end
 
-function translate!(taped::Taped, ir::IRTools.IR)
+kind(e) = Val(:nonexpr)
+kind(e::Expr) = Val(e.head)
```
devmotion, line 223: This can't be inferred - it is necessary to use `Val` here?

src/tapedfunction.jl
```diff
@@ -168,127 +154,180 @@ function (instr::Instruction{F})() where F
...
+    for (idx, line) in enumerate(ir.code)
+        isa(line, Core.Const) && (line = line.val) # unbox Core.Const
+        builder(Core.SSAValue(idx), line, kind(line))
     end
-    init_ins = Instruction(args_initializer(tape[1]), tuple(tape[1].args[2:end]...),
-                           Box{Any}(nothing), taped)
+
+    init_ins = Instruction(
+        args_initializer(taped, boxes),
+        tuple((Box{Any}(nothing) for _ in 1:taped.arity)...),
```
devmotion, line 239: Can't be inferred either (since arity is not known at compile time) but avoids splatting:
```suggestion
        ntuple(_ -> Box{Any}(nothing), taped.arity),
```

src/tapedfunction.jl
```diff
@@ -168,127 +154,180 @@ function (instr::Instruction{F})() where F
...
+const IRVar = Union{Core.SSAValue, Core.SlotNumber}
+
+struct InstructionBuilder
+    taped::Taped
+    ir::Core.CodeInfo
+    _box
+end
+
+function (builder::InstructionBuilder)(var::IRVar, line::Core.NewvarNode, ::Val)
+    (; taped, ir, _box) = builder
```
devmotion, line 253: This requires Julia 1.7. Fine in this PR but has to rewritten when support for Julia 1.6 is added.

devmotion: Same in other functions below, of course.

src/tapedfunction.jl
```diff
@@ -330,46 +367,39 @@ function Base.copy(x::Instruction, on_tape::Taped, roster::Dict{UInt64, Any})
...
+        i == 1 && continue
         new_ins = copy(x, on_tape, roster)
         new_tape[i] = new_ins
     end
 
-    init_ins = Instruction(args_initializer(new_tape[2]), tuple(new_tape[2].args[2:end]...),
-                           Box{Any}(nothing), on_tape)
+    init_ins = Instruction(
+        args_initializer(on_tape, roster),
+        tuple((Box{Any}(nothing) for _ in 1:on_tape.arity)...),
```
devmotion, line 390:
```suggestion
        ntuple(_ -> Box{Any}(nothing), on_tape.arity),
```

src/tapedfunction.jl
```diff
@@ -61,11 +54,10 @@ mutable struct TapedFunction{F} <: Taped
         tf.func = f
         tf.tape = RawTape()
         tf.counter = 1
-        tf.block_map = Dict{Int, Int}()
 
         if init # init
             tf.arity = length(args)
-            ir = IRTools.@code_ir tf.func(args...)
+            ir = code_inferred(tf.func, args_type...)
```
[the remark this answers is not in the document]
KDr2: No, `Tuple{types...}` works, but `args_type` is `::Tuple{<:DataType...}`

src/tapedfunction.jl
KDr2: Only calls with `Expr` are dispatched on `Val` now :)

rikhuijzer:
> How about letting Libtask v0.7.x (an LTS version as well?) keep using IRTools.jl and letting Julia prior to v1.7 use that version.

I think it’s a massive technical burden to maintain two versions in parallel

devmotion:
Yes, and in particular versions that are so different. E.g., in SpecialFunctions we backport all bug fixes to SpecialFunctions 1 to ensure that they also reach packages that have not updated their compat entry (since SpecialFunctions is such a core package and so many other packages depend on it), and that's easy since the breaking change in SpecialFunctions 2.0.0 was only the removal of the type piracy `factorial(::Number)`. However, if Libtask 0.7 depends on IRTools and 0.8 on CodeInfoTools, then I assume it will be much more difficult to backport bug fixes  and import performance improvements.

devmotion:
Is there a specific reason why it is difficult to support 1.6 but 1.8 works fine even though the CodeInfo is different from the one in 1.7 in both cases? In any case, I think it's fine to add support for 1.6 in a follow-up PR - but I also think it would be good to properly support, and hence also publish updates for, the LTS version.

src/tapedfunction.jl
```diff
@@ -168,127 +156,184 @@ function (instr::Instruction{F})() where F
...
         end
     end
 end
 
-function translate!(taped::Taped, ir::IRTools.IR)
+function translate!(taped::Taped, ir::Core.CodeInfo)
     tape = taped.tape
-    boxes = Dict{IRTools.Variable, Box{Any}}()
+    boxes = Dict{Symbol, Box{<:Any}}()
     _box = (x) -> arg_boxer(x, boxes)
```
devmotion, line 230: See suggestion below:
```suggestion
```

src/tapedfunction.jl
```diff
@@ -168,127 +156,184 @@ function (instr::Instruction{F})() where F
...
-                ins = BranchInstruction(
-                    cond, br.block, map(_box, br.args), taped)
-                push!(tape, ins)
-            end
-        end
+    builder = InstructionBuilder(taped, ir, _box)
+
+    for (idx, line) in enumerate(ir.code)
+        isa(line, Core.Const) && (line = line.val) # unbox Core.Const
+        builder(Core.SSAValue(idx), line)
```
devmotion, line 236: I think a simpler design would be
```suggestion
    for (idx, line) in enumerate(ir.code)
        push!(taped, _instruction(Core.SSAValue(idx), _unbox(line), taped, ir, boxes))
```
since then you could remove the `InstructionBuilder`, avoid the unpacking in every function, and could remove the `push!` in all functions.

Additionally, one would define
```julia
_unbox(line) = line
_unbox(line::Core.Const) = line.val
```

src/tapedfunction.jl
```diff
@@ -168,127 +156,184 @@ function (instr::Instruction{F})() where F
...
+    ins = GotoInstruction(cond, line.dest + 1, taped)
+    push!(taped, ins)
+end
+
+function (builder::InstructionBuilder)(var::IRVar, line::Core.GotoNode)
+    (; taped, ir, _box) = builder
+    cond = Box{Any}(false) # unify the condiftion type
+    ins = GotoInstruction(cond, line.label + 1, taped)
+    push!(taped, ins)
+end
```
devmotion, line 291:
```suggestion
function _instruction(@nospecialize(::IRVar), line::Core.GotoNode, taped::Taped, ::Core.CodeInfo, @nospecialize(boxes))
    cond = Box{Any}(false)
    return GotoInstruction(cond, line.label + 1, taped)
end
```

yebai:
> Is there a specific reason why it is difficult to support 1.6 but 1.8 works fine even though the CodeInfo is different from the one in 1.7 in both cases? 

There is a major compiler refactoring between 1.6 and 1.7. The `CodeInfo` data structures are quite similar for 1.7 and 1.8. 

> In any case, I think it's fine to add support for 1.6 in a follow-up PR - but I also think it would be good to properly support, and hence also publish updates for, the LTS version.

The old `Libtask` mechanism does support Julia 1.6 very well. The new tape-based mechanism is introduced for Julia 1.7. It is (partially) motivated by the high maintenance burden of `Libtask` for new Julia releases. It might not add too much value by backporting the new mechanism for previous Julia releases, since the old implementation works quite well.

src/tapedfunction.jl
yebai, line 236: The API is a bit confusing. Maybe consider something more explicit, e.g. 
```julia
  # builder = InstructionBuilder(taped, ir, _box)
  translate!(builder, Core.SSAValue(idx), line)
```

src/tapedfunction.jl
```diff
@@ -168,127 +156,184 @@ function (instr::Instruction{F})() where F
...
+        builder(line.args[1], rh, Val(rh.head))
+    else # rh is a single value
+        ins = Instruction(identity, (_box(rh),), _box(line.args[1]), taped)
+        push!(taped, ins)
+    end
+end
+
+function (builder::InstructionBuilder)(var, line)
+    @error "Unknown IR code: " typeof(var) var typeof(line) line
+end
```
devmotion, line 336: This will throw an error anyways:
```suggestion
```

devmotion (commented):
I think we should not even use `Val` for the `Expr` case. Unrelated of the `Expr` case, I think it is more natural to pass `ir`, `taped`, and `boxes` (not `_box`) as arguments to a function instead of constructing a callable. It seems there are also some undefined variables, so it seems not all functions are tested?

src/tapedfunction.jl
yebai, line 230: Maybe consider adding the `_fn` suffix to indicate that this is a function? 
```suggestion
    _box_fn = (x) -> arg_boxer(x, boxes)
```

Alternative, we can also introduce a new type, and turn `boxes` into a callable object, e.g. 

```julia
_box = (x) -> arg_boxer(x, boxes)

# becomes 
box_fn(x) 
```

where `box_fn` is a struct that stores the current `boxes` dictionary.

src/tapedfunction.jl
```diff
@@ -168,127 +156,184 @@ function (instr::Instruction{F})() where F
...
     end
-end
 
-function (instr::ReturnInstruction)()
-    instr.tape.retval = val(instr.arg)
+function _find_slot(all_boxes::Dict{Symbol, Box{<:Any}}, slot::Int)
+    box_id = Symbol(:_, slot)
+    return get(all_boxes, box_id) do
+        return Box{Any}(nothing) # func or argument is not used
+    end
```
yebai, line 212: Maybe consider simplifying this into: 
```suggestion
    return get(all_boxes, box_id, Box{Any}(nothing))
```

src/tapedfunction.jl
```diff
@@ -168,127 +156,184 @@ function (instr::Instruction{F})() where F
...
+arg_boxer(var, boxes::Dict{Symbol, Box{<:Any}}) = var
+arg_boxer(var::Core.SSAValue, boxes::Dict{Symbol, Box{<:Any}}) = arg_boxer(Symbol(var.id), boxes)
+arg_boxer(var::Core.TypedSlot, boxes::Dict{Symbol, Box{<:Any}}) =
+    arg_boxer(Symbol(:_, var.id), boxes)
+arg_boxer(var::Core.SlotNumber, boxes::Dict{Symbol, Box{<:Any}}) =
+    arg_boxer(Symbol(:_, var.id), boxes)
+arg_boxer(var::Symbol, boxes::Dict{Symbol, Box{<:Any}}) =
+    get!(boxes, var) do
+        return Box{Any}(var, nothing)
     end
```
yebai, line 206: Maybe consider simplifying this into 
```suggestion
    get!(boxes, var, Box{Any}(var, nothing))
```

yebai (changes requested):
Thanks, @KDr2 - excellent work. I have some minor comments below, mostly about clarity.

src/tapedfunction.jl
KDr2: But printing out more information on error is much useful to fix it. Without this, we only get a no-method matched error.

src/tapedfunction.jl
```diff
@@ -168,127 +156,184 @@ function (instr::Instruction{F})() where F
...
 end
 
+const IRVar = Union{Core.SSAValue, Core.SlotNumber}
+
+struct InstructionBuilder
+    taped::Taped
+    ir::Core.CodeInfo
+    _box
+end
+
```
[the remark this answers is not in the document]
KDr2: I prefer not to pass so many arguments here and there, that makes long function call expressions. If there are some arguments that are all the same in a bunch of function calls, it is natural to capsulate them into a context.

src/tapedfunction.jl
```diff
@@ -168,127 +156,184 @@ function (instr::Instruction{F})() where F
...
+    (; taped, ir, _box) = builder
+    ins = GotoInstruction(Box{Any}(true), 0, taped) # a noop
+    push!(taped, ins)
+end
+
+function (builder::InstructionBuilder)(var::IRVar, line::GlobalRef)
+    (; taped, ir, _box) = builder
+    ins = Instruction(val, (line,), _box(var), taped)
+    push!(taped, ins)
+end
```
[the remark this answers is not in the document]
KDr2: It's defined here: https://github.com/TuringLang/Libtask.jl/pull/126/files#diff-32dbddadf7a543bf11f421085d2833eaed85ae8599084d725af862eccd8dae2fR83

src/tapedfunction.jl
```diff
@@ -168,127 +156,184 @@ function (instr::Instruction{F})() where F
...
-        blk_ins = instr.tape[target_idx]
-        @assert isa(blk_ins, BlockInstruction)
-        @assert length(instr.args) == length(blk_ins.args)
-        for i in 1:length(instr.args)
-            blk_ins.args[i].val = val(instr.args[i])
-        end
-        instr.tape.counter = target_idx
-    else
-        instr.tape.counter += 1
+arg_boxer(var, boxes::Dict{Symbol, Box{<:Any}}) = var
```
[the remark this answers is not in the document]
KDr2: This is used for literal constants like `1`, `"a-string"`.

src/tapedfunction.jl
KDr2: @devmotion said `get!(boxes, var, Box{Any}(var, nothing))` will call `Box{Any}(var, nothing)` whether `haskey(boxes, var)` or not, and the function version will avoid some unnecessary call. I think he is right.

KDr2:
> In addition, the function names `arg_boxer` and `args_initializer` are a bit misleading.

I renamed `arg_boxer` to `var_boxer`.

src/tapedfunction.jl
```diff
@@ -168,127 +156,184 @@ function (instr::Instruction{F})() where F
...
+    (; taped, ir, _box) = builder
+    # args[1] (the left hand) is a SlotNumber, and it should be the output
+    rh = line.args[2] # the right hand, maybe a Expr, or a var, or ...
+    if Meta.isexpr(rh, [:new, :call])
+        builder(line.args[1], rh, Val(rh.head))
+    else # rh is a single value
+        ins = Instruction(identity, (_box(rh),), _box(line.args[1]), taped)
+        push!(taped, ins)
+    end
+end
```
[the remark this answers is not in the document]
devmotion: Constant propagation helps a lot:
```julia
julia> function f(expr::Expr)
           head = expr.head
           if head === :new
               return "hello"
           elseif head === :call
               return "world"
           elseif head === :(=)
               return "equals"
           elseif head === :(+)
               return "sum"
           elseif head === :(-)
               return "minus"
           elseif head === :(*)
               return "times"
           elseif head === :(/)
               return "div"
           else
               error("function not defined for head: ", head)
           end
       end

julia> g(expr::Expr) = g(expr, Val(expr.head))

julia> g(::Expr, ::Val{:new}) = "hello"
g (generic function with 2 methods)

julia> g(::Expr, ::Val{:call}) = "world"
g (generic function with 3 methods)

julia> g(::Expr, ::Val{:(=)}) = "equals"
g (generic function with 4 methods)

julia> g(::Expr, ::Val{:(+)}) = "sum"
g (generic function with 5 methods)

julia> g(::Expr, ::Val{:(-)}) = "minus"
g (generic function with 6 methods)

julia> g(::Expr, ::Val{:(*)}) = "times"
g (generic function with 7 methods)

julia> g(::Expr, ::Val{:(/)}) = "div"
g (generic function with 8 methods)

julia> using BenchmarkTools

julia> @btime f($(:(x * y)));
  4.194 ns (0 allocations: 0 bytes)

julia> @btime g($(:(x * y)));
  174.694 ns (0 allocations: 0 bytes)
```
See also: https://discourse.julialang.org/t/ann-valsplit-jl-compile-away-dynamic-dispatch-on-val-typed-arguments/77096

src/tapedfunction.jl
devmotion: True. It would better to throw an error then and not only print an error message though.

src/tapedfunction.jl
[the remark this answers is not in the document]
devmotion: Ah I see. Then the suggestion is
```suggestion
function _instruction(var::IRVar, line::GlobalRef, taped::Taped, ::Core.CodeInfo, boxes)
    return Instruction(val, (line,), arg_boxer(var, boxes), taped)
end
```

src/tapedfunction.jl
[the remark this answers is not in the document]
devmotion: I think the mental overhead of the `InstructionBuilder` outweighs the slightly shorter function signatures. The function is also only called in one place, so there are not many lengthy calls.

An additional advantage of passing the arguments is that you can mark a subset of them with `@nospecialize` to avoid specialization on them if they are not used in the function body.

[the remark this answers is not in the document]
devmotion: In any case it seems simpler and easier to read if we
- only return the instruction and `push!` inside the for loop
- pass `boxes` around instead of a `_box` closure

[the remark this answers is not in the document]
KDr2: - [x] only return the instruction and push! inside the for loop
- [x] pass boxes around instead of a _box closure

But the first function signature below is too long for me, let's keep the builder for now...

```
function _instruction(taped::Taped, ir::Core.CodeInfo, boxes::Dict{Symbol, Box{:<Any}}, var::IRVar, line::Core.NewvarNode)
function (builder::InstructionBuilder)(var::IRVar, line::Core.NewvarNode)
```

Issue #127, opened 2022-02-23
Remove the dependency on `MacroTools`


KDr2:
We only use `MacroTools.@forward` on `TapedFunction` now, so it can be removed easily.

EDIT:
It's not `now`, it's after https://github.com/TuringLang/Libtask.jl/pull/126 is merged.

Issue #97, opened 2021-12-13
Seprate code and data in instructions 


yebai:
The current `Instruction` type 

```julia
mutable struct Instruction{F}
    fun::F
    input::Tuple
    output
    tape
end
```

carries both the code (i.e. `fun::F`) and data (i.e. `input` and `output`). If we want to support loops suggested [here](https://github.com/TuringLang/Libtask.jl/issues/96), we likely will need a mechanism to store multiple copies of data in the same instruction. It is because each instruction might run multiple times assuming we copy the control flow IR onto the tape without linearization. Maybe we can turn `input` and `output` into a vector? 

```julia
struct Instruction{F}
    fun::F
    io::Vector{InstructionIO}
    tape
end

struct InstructionIO
    input::Tuple
    output
end
```

KDr2:
Running an instruction multiple times doesn't interfere with data copying I think, so maybe we need not do this?

PR #128, opened 2022-03-03
Separate data and instruction

KDr2:
Implement #127 and #97.

The type info task will be in another PR.

### DIFF
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -8,10 +8,9 @@ Box(x) = Box(gensym(), x)
 Box{T}(x) where {T} = Box{T}(gensym(), x)
 Base.show(io::IO, box::Box) = print(io, "Box(", box.id, ")")
 
-## Instruction and Taped
+## Instruction and TapedFunction
 
 abstract type AbstractInstruction end
-abstract type Taped end
 const RawTape = Vector{AbstractInstruction}
 
 """
@@ -19,30 +18,30 @@ const RawTape = Vector{AbstractInstruction}
 
 An `Instruction` stands for a function call
 """
-mutable struct Instruction{F, TI<:Tuple, TO} <: AbstractInstruction
+mutable struct Instruction{F, N} <: AbstractInstruction
     func::F
-    input::TI
-    output::Box{TO}
+    input::NTuple{N, Symbol}
+    output::Symbol
 end
 
 mutable struct GotoInstruction <: AbstractInstruction
-    condition::Box{Any}
+    condition::Symbol
     # we enusre a 1-to-1 mapping between ir.code and instruction
-    # so here we can use the index directly (actually index+1, because
-    # we have an extra args_initializer instruction at the beginning of the tape).
+    # so here we can use the index directly.
     dest::Int
 end
 
-mutable struct ReturnInstruction{TA} <: AbstractInstruction
-    arg::TA
+mutable struct ReturnInstruction <: AbstractInstruction
+    arg::Symbol
 end
 
-mutable struct TapedFunction{F} <: Taped
+mutable struct TapedFunction{F}
     func::F # maybe a function, a constructor, or a callable object
     arity::Int
     ir::Core.CodeInfo
     tape::RawTape
     counter::Int
+    bindings::Dict{Symbol, Box{<:Any}}
     retval
     owner
 
@@ -78,18 +77,24 @@ const TRCache = LRU{Tuple, TapedFunction}(maxsize=10)
 
 val(x) = x
 val(x::Box) = x.val
+val(x::Box{GlobalRef}) = val(x.val)
+val(x::Box{QuoteNode}) = val(x.val)
 val(x::GlobalRef) = getproperty(x.mod, x.name)
 val(x::QuoteNode) = eval(x)
 val(x::TapedFunction) = x.func
 result(t::TapedFunction) = t.retval
 
 function (tf::TapedFunction)(args...; callback=nothing)
-    # run the raw tape
-    if(tf.counter <= 1 && length(args) > 0)
-        input = map(Box{Any}, args)
-        tf.tape[1].input = input
+    # set args
+    if(tf.counter <= 1)
+        haskey(tf.bindings, :_1) && (tf.bindings[:_1].val = tf.func)
+        for i in 1:length(args)
+            slot = Symbol("_", i + 1)
+            haskey(tf.bindings, slot) && (tf.bindings[slot].val = args[i])
+        end
     end
 
+    # run the raw tape
     while true
         ins = tf.tape[tf.counter]
         ins(tf)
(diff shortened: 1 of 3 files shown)


KDr2:
This PR reduces the Integrate Test from 140-160 mins to about 100 mins.

rikhuijzer:
> This PR reduces the Integrate Test from 140-160 mins to about 100 mins.

Tested locally or on the runners?

src/tapedfunction.jl
```diff
@@ -142,34 +143,48 @@ function Base.show(io::IO, instr::Instruction)
...
+function (instr::Instruction{F})(tf::TapedFunction) where F
     # catch run-time exceptions / errors.
     try
-        func = val(instr.func)
-        output = func(map(val, instr.input)...)
-        instr.output.val = output
-        taped.counter += 1
+        func = val(_lookup(tf, instr.func))
+        input_boxes = map(x -> _lookup(tf, x), instr.input)
+        output = func(map(val, input_boxes)...)
```
devmotion, line 158:
```suggestion
        inputs = map(x -> val(_lookup(tf, x)), instr.input)
        output = func(inputs...)
```

src/tapedfunction.jl
```diff
@@ -191,38 +206,21 @@ end
 
 ## Translation: CodeInfo -> Tape
 
-var_boxer(var, boxes::Dict{Symbol, Box{<:Any}}) = var # for literal constants
+var_boxer(var, boxes::Dict{Symbol, Box{<:Any}}) = begin # for literal constants
```
devmotion, line 209:
```suggestion
function var_boxer(var, boxes::Dict{Symbol, Box{<:Any}})
    # for literal constants
```

src/tapedfunction.jl
```diff
@@ -142,34 +143,48 @@ function Base.show(io::IO, instr::Instruction)
...
-    if val(instr.condition)
-        taped.counter += 1
+function (instr::GotoInstruction)(tf::TapedFunction)
+    cond = if instr.condition === :_true
+        true
+    elseif  instr.condition === :_false
+        false
+    else
+        val(_lookup(tf, instr.condition))
+    end
```
rikhuijzer, line 177:
```suggestion
    cond = instr.condition === :_true ? true :
        instr.condition === :_false ? false :
        val(_lookup(tf, instr.condition))
```

rikhuijzer (commented):
I don't understand most of the code, but I like that the struct fields have clearer types

src/tapedfunction.jl
```diff
@@ -57,45 +56,47 @@ mutable struct TapedFunction{F} <: Taped
...
-    if(tf.counter <= 1 && length(args) > 0)
-        input = map(Box{Any}, args)
-        tf.tape[1].input = input
+    # set args
+    if(tf.counter <= 1)
+        haskey(tf.bindings, :_1) && (tf.bindings[:_1].val = tf.func)
+        for i in 1:length(args)
+            slot = Symbol("_", i + 1)
+            haskey(tf.bindings, slot) && (tf.bindings[slot].val = args[i])
+        end
```
devmotion, line 94: It would be faster if one could avoid looking up every key twice, once in `haskey` and once in `setindex!`.

One alternative would be to use `get` with a default value such as a  `NoKeyFound()` singleton instead of `haskey` (`nothing` seems ambiguous for the arguments unfortunately). Another alternative would be to iterate over keys - but this would only work, or be efficient at least, if most keys are `:_i` with <= `length(args) + 1`.

KDr2:
> > This PR reduces the Integrate Test from 140-160 mins to about 100 mins.
> 
> Tested locally or on the runners?

I simply read the time spent on github action, but it may be not so accurate due to the job concurrency and queue?

src/tapedfunction.jl
KDr2: We need an `Option` or `Maybe` type in Julia 😄 

Here most keys in the Dict will not be `:_i`.

KDr2:
> > > This PR reduces the Integrate Test from 140-160 mins to about 100 mins.
> > 
> > 
> > Tested locally or on the runners?
> 
> I simply read the time spent on github action, but it may be not so accurate due to the job concurrency and queue?

You can see the time the IntegerateTest consumed here: https://github.com/TuringLang/Libtask.jl/actions/workflows/IntegrationTest.yml

rikhuijzer:
> This PR reduces the Integrate Test from 140-160 mins to about 100 mins.

Nice. I can confirm that this PR is good for performance. I ran the following subset of the Turing tests (`Turing.jl/test/runtests.jl`) on my system:

```julia
[...]
include(pkgdir(Turing)*"/test/test_utils/AllUtils.jl")

using Logging
Logging.disable_logging(Logging.Info)

@testset "Turing" begin
    @time include("inference/gibbs.jl")
    @time include("contrib/inference/sghmc.jl")
    @time include("stdlib/RandomMeasures.jl")
end
```

### `master` branch 
```
 43.139966 seconds (87.17 M allocations: 5.339 GiB, 3.38% gc time, 94.57% compilation time)
443.315047 seconds (1.18 G allocations: 77.239 GiB, 3.21% gc time, 18.66% compilation time)
 13.570206 seconds (39.32 M allocations: 2.376 GiB, 4.77% gc time, 98.35% compilation time)
  9.087712 seconds (17.42 M allocations: 1.047 GiB, 4.29% gc time, 73.16% compilation time)
```

### `sep-data` branch (this PR)
```
 41.664786 seconds (84.94 M allocations: 5.297 GiB, 3.60% gc time, 94.48% compilation time)
390.978920 seconds (842.42 M allocations: 69.394 GiB, 3.04% gc time, 20.74% compilation time)
 13.259697 seconds (39.32 M allocations: 2.376 GiB, 4.87% gc time, 98.34% compilation time)
  8.926231 seconds (17.57 M allocations: 1.072 GiB, 4.81% gc time, 67.26% compilation time)
```

rikhuijzer:
> > > This PR reduces the Integrate Test from 140-160 mins to about 100 mins.
> > 
> > 
> > Tested locally or on the runners?
> 
> I simply read the time spent on github action, but it may be not so accurate due to the job concurrency and queue?

Mostly due to different CPUs. Not all runners use the same CPU and the difference in running time can be as big as 40%.

KDr2:
> Nice. I can confirm that this PR is good for performance.

I think the main reason behind this is that we now only need to copy the data(variables), and don't need to copy the instructions anymore.

src/tapedfunction.jl
```diff
@@ -8,45 +8,44 @@ Box(x) = Box(gensym(), x)
 Box{T}(x) where {T} = Box{T}(gensym(), x)
 Base.show(io::IO, box::Box) = print(io, "Box(", box.id, ")")
 
-## Instruction and Taped
+## Instruction and TapedFunction
 
 abstract type AbstractInstruction end
-abstract type Taped end
```
yebai, line 14: Maybe keep this type for now - it’ll become useful again in future work.

src/tapedfunction.jl
```diff
@@ -8,45 +8,44 @@ Box(x) = Box(gensym(), x)
...
+mutable struct TapedFunction{F}
     func::F # maybe a function, a constructor, or a callable object
     arity::Int
     ir::Core.CodeInfo
     tape::RawTape
     counter::Int
-    retval
+    bindings::Dict{Symbol, Box{<:Any}}
+    retval::Symbol
     owner
```
yebai, line 46: Do we still need this `owner` field?

src/tapedfunction.jl
```diff
@@ -57,45 +56,47 @@ mutable struct TapedFunction{F} <: Taped
...
-            TRCache[cache_key] = tf
-        end
+        translate!(tf, ir)
+        TRCache[cache_key] = tf # set cache
         return tf
     end
+
+    function TapedFunction(tf::TapedFunction{F}) where {F}
+        new{F}(tf.func, tf.arity, tf.ir, tf.tape,
+               tf.counter, tf.bindings, :none, tf.owner)
```
yebai, line 72: Should `:none` be `tf.retval` here?

src/tapedfunction.jl
yebai: The plan is to replace `to.bindings::Dict` with `NamedTuple` in the next PR. So it probably won’t matter.

src/tapedfunction.jl
```diff
@@ -244,12 +234,12 @@ function translate!!(var::IRVar, line::Core.NewvarNode,
...
     # use a noop to ensure the 1-to-1 mapping from ir.code to instructions
     # on tape. see GotoInstruction.dest.
-    return GotoInstruction(Box{Any}(true), 0)
+    return GotoInstruction(:_true, 0)
 end
 
 function translate!!(var::IRVar, line::GlobalRef,
                      boxes::Dict{Symbol, Box{<:Any}}, @nospecialize(ir))
-    return Instruction(val, (line,), var_boxer(var, boxes))
+    return Instruction(() -> val(line), (), var_boxer(var, boxes))
```
yebai, line 242: Maybe keep the previous implementation since it avoids creating unnecessary anymous function? 
```suggestion
    return Instruction(val, (line,), var_boxer(var, boxes))
```

yebai (changes requested):
Thanks @KDr2 - nice improvements!

src/tapedfunction.jl
KDr2: Generally, here `line` is not a Symbol, it may be a `GlobalRef` in most cases, so it can't be put into `Instruction.input` which is a tuple of symbols.

src/tapedfunction.jl
KDr2: tf.retval is `undef` when the function is not terminated (no `ReturnInstruction` had been executed), we can't access it at the moment, so I use `:none` instead. It has the same effect as `undef`: one will get an error if one calls `result(tf)` when it hasn't returned.

src/tapedfunction.jl
KDr2: We don't if we will not switch to https://github.com/TuringLang/Libtask.jl/blob/master/src/tapedtask.jl#L76

src/tapedfunction.jl
KDr2: At this moment, it only has one subtype: TapedFunction, so I removed it. And I think it's not hard to add it back when we need it.

But I'm OK to take it back now if you insist to do so 😄

src/tapedfunction.jl
KDr2: I removed it.

yebai (approved):

yebai:
I think it’s ready to merge when the CI passes.

PR #130, opened 2022-03-05
Trying some other optimizations beside type-infer

KDr2:


KDr2:
```
@testset "Turing" begin
    @time include("inference/gibbs.jl")
    @time include("contrib/inference/sghmc.jl")
    @time include("stdlib/RandomMeasures.jl")
end
```

Branch master:

```
 34.395940 seconds (82.78 M allocations: 5.174 GiB, 4.03% gc time, 95.12% compilation time)
285.165249 seconds (834.03 M allocations: 67.579 GiB, 3.62% gc time, 25.78% compilation time)
 11.141108 seconds (38.60 M allocations: 2.341 GiB, 4.64% gc time, 98.42% compilation time)
  7.159114 seconds (16.94 M allocations: 1.029 GiB, 4.85% gc time, 70.14% compilation time)
```

Branch no-type-opt:
```
 39.743907 seconds (89.76 M allocations: 5.518 GiB, 3.81% gc time, 95.43% compilation time)
299.946079 seconds (831.72 M allocations: 67.493 GiB, 3.45% gc time, 24.85% compilation time)
 11.235855 seconds (38.59 M allocations: 2.341 GiB, 4.98% gc time, 97.71% compilation time)
  9.119536 seconds (26.03 M allocations: 1.275 GiB, 4.52% gc time, 75.07% compilation time)
```

It becomes slower... 😢

KDr2:
First commit(5ac43d4) on Branch no-type-opt:
```
 34.459184 seconds (81.37 M allocations: 5.088 GiB, 3.99% gc time, 94.95% compilation time)
302.663785 seconds (553.19 M allocations: 59.879 GiB, 3.18% gc time, 23.32% compilation time)
 10.996351 seconds (38.60 M allocations: 2.341 GiB, 5.07% gc time, 98.22% compilation time)
  7.437808 seconds (15.85 M allocations: 1012.160 MiB, 4.32% gc time, 69.07% compilation time)
```

KDr2:
Second commit(cce6cb3) on Branch no-type-opt:
```
 34.994963 seconds (82.41 M allocations: 5.142 GiB, 4.03% gc time, 95.08% compilation time)
300.684379 seconds (615.39 M allocations: 62.235 GiB, 3.44% gc time, 23.27% compilation time)
 10.994031 seconds (38.60 M allocations: 2.341 GiB, 4.99% gc time, 98.45% compilation time)
  7.422173 seconds (17.95 M allocations: 1.065 GiB, 4.46% gc time, 69.01% compilation time)
```

KDr2:
> First commit([5ac43d4](https://github.com/TuringLang/Libtask.jl/commit/5ac43d4515343832b16ab00a41953cd9a161d1da)) on Branch no-type-opt:

This commit has the least memory allocation, but it doesn't save much time cost.

rikhuijzer:
> > First commit([5ac43d4](https://github.com/TuringLang/Libtask.jl/commit/5ac43d4515343832b16ab00a41953cd9a161d1da)) on Branch no-type-opt:
> 
> This commit has the least memory allocation, but it doesn't save much time cost.

Could be that your computer was more busy during the one run versus the other. What you can do to dive deeper into why things are going faster or slower is to put `@time` things at some places in the codebase. As far as I can tell, it's difficult to see these instruction constructors as stand-alone, so then running larger parts with a `@time` does help sometimes.

src/tapedfunction.jl
```diff
@@ -1,48 +1,43 @@
...
     dest::Int
 end
 
-struct ReturnInstruction <: AbstractInstruction
-    arg::Symbol
+struct ReturnInstruction{T} <: AbstractInstruction
+    arg::TypedVar{T}
 end
 
 mutable struct TapedFunction{F}
```
rikhuijzer, line 32:
```suggestion
mutable struct TapedFunction{F,T}
```

src/tapedfunction.jl
```diff
@@ -1,48 +1,43 @@
...
     func::F # maybe a function, a constructor, or a callable object
     arity::Int
     ir::Core.CodeInfo
     tape::RawTape
+    unified_tape::Vector{FunctionWrapper{Nothing, Tuple{TapedFunction}}}
     counter::Int
-    bindings::Dict{Symbol, Box{<:Any}}
-    retval::Symbol
+    bindings::Dict{Symbol, Any}
+    retval::TypedVar
```
rikhuijzer, line 40:
```suggestion
    retval::TypedVar{T}
```
This makes `retval` a concrete type again which avoids dynamic dispatch. The original `retval::Symbol` is more performant.

src/tapedfunction.jl
```diff
@@ -201,116 +192,121 @@ end
...
 
-function var_boxer(var, boxes::Dict{Symbol, Box{<:Any}}) # for literal constants
-    box = Box(var)
-    boxes[box.id] = box
-    return box.id
+function bind_var!(var, bindings::Dict{Symbol, Any}, ir::Core.CodeInfo) # for literal constants
+    id = gensym()
+    bindings[id] = var
+    return TypedVar{typeof(var)}(id)
+end
```
rikhuijzer, line 199:
```suggestion
function bind_var!(var::T, bindings::Dict{Symbol, Any}, ir::Core.CodeInfo) where {T} # for literal constants
    id = gensym()
    bindings[id] = var
    return TypedVar{T}(id)
end
```

rikhuijzer (commented):
I don't understand everything. Some suggestions.

Good news about the benchmarking: https://github.com/TuringLang/Turing.jl/pull/1787 is merged, so we can now see the allocations in the integration test.

src/tapedfunction.jl
devmotion: Is there a specific reason for the static parameter? The general style guide suggests using them only if necessary: https://docs.julialang.org/en/v1/manual/style-guide/index.html#Don't-use-unnecessary-static-parameters

KDr2: You mean `T` in `TypedVar{T}`? The idea is, every var is a Symbol, but we don't know its type, so when executing the instruction, `input= _look_up(var_name) ;instr.func(input)`, the type of `input` is unknown. Here I try to change a var from a symbol to a `TypedVar{T}` in which `T` is its type and `id` is its name. So I expect with `input = _look_up(typedvar)::T` the compiler will know the input's type. I am not really sure if this helps, from the benchmark result I should see it doesn't help.

rikhuijzer: Ah okay so my suggestion is wrong @devmotion? Julia will dispatch on `var` anyway and then `typeof(var)` will be inlined. I'll resolve this conversation then.

src/tapedfunction.jl
```diff
@@ -147,16 +149,17 @@ end
 
 
 _lookup(tf::TapedFunction, v) = v
-_lookup(tf::TapedFunction, v::Symbol) = tf.bindings[v]
+_lookup(tf::TapedFunction, v::TypedVar{T}) where T = tf.bindings[v.id]::T
```
yebai, line 152: @KDr2 this function is probably not type-stable, despite the type assertion. The reason is that type assertion does not help with type inference (i.e. at compile time, we can't infer type of `tf.bindings[v.id]`). But rather, type assertion check the type of `tf.bindings[v.id]` at runtime, and throw an error if type constraint is violated. It should be possible to check this with a simplified example. This would also explain why the performance does not improve much with this PR.

yebai: See e.g. https://docs.julialang.org/en/v1/manual/performance-tips/#Annotate-values-taken-from-untyped-locations

> Type annotation will not enhance (and can actually hinder) performance if the type is abstract, or constructed at run-time.

src/tapedfunction.jl
```diff
@@ -184,9 +188,18 @@ end
 
 
 ## internal functions
+function fix_ir(ir::Core.CodeInfo)
+    if Base.JLOptions().code_coverage != 0 &&
```
rikhuijzer, line 192: What does this do?

KDr2: See https://github.com/JuliaCompilerPlugins/CodeInfoTools.jl/issues/18, but I found this function doesn't fix it in all cases, I will revert or fix it soon...

src/tapedfunction.jl
```diff
@@ -56,72 +25,121 @@ mutable struct TapedFunction{F}
...
+    id::Symbol
+    get # ::FunctionWrapper{T, Tuple{TapedFunction, Symbol}}
+    set # ::FunctionWrapper{Nothing, Tuple{TapedFunction, Symbol, T}}
+
+    function Box{T}(id::Symbol) where T
+        return new(id, _inner_getter, _inner_setter)
+        # return new(id,
+        # FunctionWrapper{T, Tuple{TapedFunction, Symbol}}(_inner_getter),
+        # FunctionWrapper{Nothing, Tuple{TapedFunction, Symbol, T}}(_inner_setter))
+    end
```
KDr2, line 63: @yebai here if I use `FunctionWrapper` (the commented lines), I will get many errors related to types and compilers.

Maybe `FunctionWrapper` has some glitches on Julia 1.7.

src/tapedfunction.jl
```diff
@@ -56,72 +25,132 @@ mutable struct TapedFunction{F}
...
 
+function unify!(tf::TapedFunction)
+    length(tf.tape) == length(tf.unified_tape) && return
+    for ins in tf.tape
+        push!(tf.unified_tape, FunctionWrapper{Nothing, Tuple{TapedFunction}}(ins))
+    end
+end
+
+# const TypedFunction = FunctionWrapper
+mutable struct TypedFunction{OT, IT<:Tuple}
```
yebai, line 54:
```suggestion
struct TypedFunction{OT, IT<:Tuple}
```

src/tapedfunction.jl
```diff
@@ -56,72 +25,132 @@ mutable struct TapedFunction{F}
...
+    func::Function
+    retval::OT
```
yebai, line 56:
```suggestion
    retval::Ref{OT}
```

src/tapedfunction.jl
```diff
@@ -56,72 +25,132 @@ mutable struct TapedFunction{F}
...
+
+    function Box{T}(id::Symbol) where T
+        return new(id,
+                   TypedFunction{T, Tuple{TapedFunction, Symbol}}(_inner_getter),
+                   TypedFunction{Nothing, Tuple{TapedFunction, Symbol, T}}(_inner_setter))
+    end
+end
+
+_inner_getter(tf::TapedFunction, v::Symbol) = tf.bindings[v]
+_inner_setter(tf::TapedFunction, v::Symbol, c) = tf.bindings[v] = c
```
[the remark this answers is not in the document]
yebai: Maybe avoid specialising in types of `c` to reduce compilation time.

yebai:
@KDr2 Maybe we should introduce a few benchmark examples that do not depend on Turing. It allows us to better understand the performance bottleneck -- the Turing integration tests are helpful, but it involves many other parts, which make the runtime numbers hard to analyse.

src/tapedfunction.jl
```diff
@@ -145,18 +174,13 @@ function Base.show(io::IO, instr::GotoInstruction)
...
 
-
-_lookup(tf::TapedFunction, v) = v
-_lookup(tf::TapedFunction, v::Symbol) = tf.bindings[v]
-
 function (instr::Instruction{F})(tf::TapedFunction) where F
     # catch run-time exceptions / errors.
     try
         func = val(_lookup(tf, instr.func))
         inputs = map(x -> val(_lookup(tf, x)), instr.input)
```
yebai, line 181: we should replace all `_lookup` and `_update_var!` calls with `box.getter`/`box.setter` in instructions.

KDr2: I keep `_lookup` because `Instruction.func` may be or may not be a `Box`. And, sometimes we also need to update a variable with a symbol but not a box, e.g., while we are initializing the arguments.

yebai: ok, that makes sense.

src/tapedfunction.jl
```diff
@@ -1,47 +1,16 @@
...
-struct ReturnInstruction <: AbstractInstruction
-    arg::Symbol
-end
-
 mutable struct TapedFunction{F}
     func::F # maybe a function, a constructor, or a callable object
     arity::Int
     ir::Core.CodeInfo
     tape::RawTape
+    unified_tape::Vector{FunctionWrapper{Nothing, Tuple{TapedFunction}}}
```
yebai, line 11: Maybe merge this `unified_tape` with `tape` by using a parametric type for `tape::T`. Then we can create a `compile(TapedFunction)` API to convert TapedFunctions from `RawTape` to `Vector{FunctionWrapper{Nothing, Tuple{TapedFunction}}}`.

src/tapedfunction.jl
```diff
@@ -56,72 +25,132 @@ mutable struct TapedFunction{F}
...
     function TapedFunction(tf::TapedFunction{F}) where {F}
-        new{F}(tf.func, tf.arity, tf.ir, tf.tape,
+        new{F}(tf.func, tf.arity, tf.ir, tf.tape, tf.unified_tape,
                tf.counter, tf.bindings, :none)
     end
 end
 
 const TRCache = LRU{Tuple, TapedFunction}(maxsize=10)
 
+function unify!(tf::TapedFunction)
```
yebai, line 46:
```suggestion
function compile(tf::TapedFunction)
```

yebai (commented):
thanks, @KDr2  - it looks good to me. I left some minor comments below. 

I think we still need some benchmarks before merging. We need to ensure each `Instruction` in a `TapedFunction` is type-sable, if the original function `f` underlying `TapedFunction` is type stabe.

src/tapedfunction.jl
```diff
@@ -56,72 +25,132 @@ mutable struct TapedFunction{F}
...
+    length(tf.tape) == length(tf.unified_tape) && return
+    for ins in tf.tape
+        push!(tf.unified_tape, FunctionWrapper{Nothing, Tuple{TapedFunction}}(ins))
+    end
+end
+
+# const TypedFunction = FunctionWrapper
+struct TypedFunction{OT, IT<:Tuple}
+    func::Function
+    retval::Ref{OT}
```
devmotion, line 56: This should be changed to
```suggestion
    retval::Base.RefValue{OT}
```
since
```julia
julia> isconcretetype(Ref{Float64})
false

julia> isconcretetype(Base.RefValue{Float64})
true
```
This makes a huge difference in practice, e.g.
```julia
julia> using BenchmarkTools

julia> struct A{T}
           x::Ref{T}
       end

julia> struct B{T}
           x::Base.RefValue{T}
       end

julia> function f(x::Union{A{T},B{T}}) where {T}
           s = zero(T)
           for _ in 1:10_000
               s += x.x[]
           end
           s
       end

julia> @btime f($(A(Ref(1.0))));
  130.273 μs (20000 allocations: 312.50 KiB)

julia> @btime f($(B(Ref(1.0))));
  8.492 μs (0 allocations: 0 bytes)
```

src/tapedfunction.jl
```diff
@@ -56,72 +25,132 @@ mutable struct TapedFunction{F}
...
+function unify!(tf::TapedFunction)
+    length(tf.tape) == length(tf.unified_tape) && return
+    for ins in tf.tape
+        push!(tf.unified_tape, FunctionWrapper{Nothing, Tuple{TapedFunction}}(ins))
+    end
+end
+
+# const TypedFunction = FunctionWrapper
+struct TypedFunction{OT, IT<:Tuple}
+    func::Function
```
devmotion, line 55: I guess this abstract type is desired?

src/tapedfunction.jl
```diff
@@ -56,72 +25,132 @@ mutable struct TapedFunction{F}
...
+    retval::Ref{OT}
+    TypedFunction{OT, IT}(f::Function) where {OT, IT<:Tuple} = new{OT, IT}(f, Ref{OT}())
+end
+function (f::TypedFunction{OT, IT})(args...) where {OT, IT<:Tuple}
+    output = f.func(args...)
+    OT === Nothing ? (f.retval[] = nothing) : (f.retval[] = output)
+    return f.retval[]
```
devmotion, line 62: A bit simpler, without accessing `retval` twice:
```suggestion
    f.reval[] = OT === Nothing ? nothing : output
```

devmotion (changes requested):
My main suggestion is to use `Base.RefValue` instead of `Ref`.

It is unclear to me if/how this affects performance. It would be good to perform some benchmarks (or (re-)post them, if they exist - I assumed the comments above are outdated).

src/tapedfunction.jl
KDr2: I didn't put it into type parameters due to that I want `TypedFunction` and `FunctionWrapper` to have the same type signature. Maybe we can ignore the usage of `FunctionWrapper` here because it is problematic and we may not use it anymore?

Issue #132, opened 2022-03-18
Add standalone benchmarks for `TapedFunction`


yebai:
We might be able to adapt the benchmark examples in `ReverseDiff` for `TapedFunction`. The only difference between `ReverseDiff.Tape` and `Libtask.TapedFunction` is that we don't do backward pass yet. But these examples can be used for benchmarking both forward pass (e.g. `TapedFunction` and executing original function) too.

PR #133, opened 2022-03-23
Add benchmarks

KDr2:
#132

### DIFF
--- a/perf/benchmark.jl
+++ b/perf/benchmark.jl
@@ -1,0 +1,87 @@
+using Libtask
+using LinearAlgebra
+using BenchmarkTools
+
+####################################################################
+
+function benchmark_driver!(f, x...)
+    println("benchmarking $(f)...")
+    tf = Libtask.TapedFunction(f, x)
+
+    print("  Run Original Function:")
+    @btime $f($(x)...)
+    GC.gc()
+
+    print("  Run TapedFunction:")
+    @btime $tf($(x)...)
+    GC.gc()
+
+    ctf = Libtask.compile(tf)
+    print("  Run TapedFunction (compiled):")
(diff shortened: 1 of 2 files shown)


perf/benchmark.jl
```diff
@@ -0,0 +1,85 @@
+using Libtask
+using LinearAlgebra
+using BenchmarkTools
+
+####################################################################
+
+function benchmark_driver!(f, x...)
```
yebai, line 7:
```suggestion
function benchmark_driver!(f, x...; f_displayname=string(f))
```

perf/benchmark.jl
```diff
@@ -0,0 +1,85 @@
...
+    println("benchmarking $(f)...")
```
yebai, line 8:
```suggestion
    println("benchmarking $(f_displayname)...")
```

perf/benchmark.jl
```diff
@@ -0,0 +1,85 @@
...
+        a = reshape(x[1:n^2], n, n)
+        b = reshape(x[n^2 + 1:2n^2], n, n)
+        return log.((a * b) + a - b)
+    end
+end
+
+n = 100
+matrix_test = generate_matrix_test(n)
+x = collect(1.0:(2n^2 + n))
+benchmark_driver!(matrix_test, x)
```
yebai, line 68:
```suggestion
benchmark_driver!(matrix_test, x; f_displayname="matrix_test")
```

yebai:
@KDr2 A few (quick) observations: 

1.  It seems the `matrix_test` is nearly 10 times slower compared to the original function. 
2. Compiled function always allocate (one) more, and about 2x slower.

PR #134, opened 2022-03-25
Benchmark `produce`

yebai:


perf/benchmark.jl
```diff
@@ -20,30 +22,49 @@ function benchmark_driver!(f, x...; f_displayname=string(f))
...
 ####################################################################
 
 
-function rosenbrock(x)
+function rosenbrock(x, callback=nothing)
     i = x[2:end]
     j = x[1:end-1]
-    return sum((1 .- j).^2 + 100*(i - j.^2).^2)
+    ret = sum((1 .- j).^2 + 100*(i - j.^2).^2)
+    if callback !== nothing; callback(ret); end
```
rikhuijzer, line 50: This is not idiomatic Julia code.

```suggestion
    callback !== nothing && callback(ret)
```
or multiple lines

perf/benchmark.jl
```diff
@@ -20,30 +22,49 @@ function benchmark_driver!(f, x...; f_displayname=string(f))
...
-function ackley(x::AbstractVector)
+function ackley(x::AbstractVector, callback=nothing)
     a, b, c = 20.0, -0.2, 2.0*π
     len_recip = inv(length(x))
     sum_sqrs = zero(eltype(x))
     sum_cos = sum_sqrs
     for i in x
         sum_cos += cos(c*i)
         sum_sqrs += i^2
+        if callback !== nothing; produce(sum_sqrs); end
```
rikhuijzer, line 67:
```suggestion
        callback !== nothing && produce(sum_sqrs)
```

perf/benchmark.jl
```diff
@@ -54,11 +75,13 @@ benchmark_driver!(ackley, x)
...
 ####################################################################
 function generate_matrix_test(n)
-    return x -> begin
+    return (x, callback=nothing) -> begin
         # @assert length(x) == 2n^2 + n
         a = reshape(x[1:n^2], n, n)
         b = reshape(x[n^2 + 1:2n^2], n, n)
-        return log.((a * b) + a - b)
+        ret = log.((a * b) + a - b)
+        if callback !== nothing;  callback(ret); end
```
rikhuijzer, line 83:
```suggestion
        callback !== nothing && callback(ret)
```

perf/benchmark.jl
```diff
@@ -71,15 +94,17 @@ benchmark_driver!(matrix_test, x; f_displayname="matrix_test")
 relu(x) = log.(1.0 .+ exp.(x))
 sigmoid(n) = 1. / (1. + exp(-n))
 
-function neural_net(w1, w2, w3, x1)
+function neural_net(w1, w2, w3, x1, callback=nothing)
     x2 = relu(w1 * x1)
     x3 = relu(w2 * x2)
-    return sigmoid(LinearAlgebra.dot(w3, x3))
+    ret = sigmoid(LinearAlgebra.dot(w3, x3))
+    if callback !== nothing; callback(ret); end
```
rikhuijzer, line 101:
```suggestion
    callback !== nothing && callback(ret)
```

perf/benchmark.jl
```diff
@@ -20,30 +22,49 @@ function benchmark_driver!(f, x...; f_displayname=string(f))
     print("  Run TapedFunction (compiled):")
     @btime $ctf($(x)...)
     GC.gc()
+
+    print("  Run TapedTask: ")
+    x = (x[1:end-1]..., produce);
+    # show the number of produce calls inside `f`
+    tf = Libtask.TapedFunction(f, x...);
+    tt = TapedTask(tf, x...); 
+    c = 0; while consume(tt)!==nothing c+=1 end;
```
rikhuijzer, line 31: Also non-idiomatic Julia code.

```suggestion
    c = 0
    while consume(tt) !== nothing
        c += 1
    end
```

perf/benchmark.jl
```diff
@@ -20,30 +22,49 @@ function benchmark_driver!(f, x...; f_displayname=string(f))
...
+    print("#produce=", c);
+    f_task = (tf, x) -> begin 
+        # tf = Libtask.TapedFunction(f, x...);
+        tt = TapedTask(tf, x...); 
+        while consume(tt)!==nothing
+        end;
```
rikhuijzer, line 37: Is this just about consuming all elements?

rikhuijzer (commented):
I've added a few suggestions to make the code more idiomatic. On top of that, the code contains a lot of semicolons at line endings. Would be good to remove those too

perf/benchmark.jl
yebai: that's correct - it is meant to consume all produced values from `tt`.

yebai:
> I've added a few suggestions to make the code more idiomatic. On top of that, the code contains a lot of semicolons at line endings. Would be good to remove those too

thanks @rikhuijzer - this PR contains a lot of code copy-and-pasted from terminals when I was debugging #135, which is a bit messy.

rikhuijzer:
In general, I like this PR by the way a lot. I had a lot of trouble debugging the functionality when I was working on Libtask. One question: do you know how well these estimates represent the workload under normal load?

KDr2 (approved):

yebai:
> One question: do you know how well these estimates represent the workload under normal load?

These examples cover loops and produce calls, which I suspect are the bottlenecks for optimal performance. The former is very sensitive to type-instability, while the latter tests the efficiency of the most frequent `TapedTask` operation. The `TArray` stuff is potentially a bottleneck for performance, but they can be depreciated now since the new tape-based mechanism provides a more flexible solution for controlling which variables get copied.

PR #138, opened 2022-04-02
Peformance Optimization

KDr2:
The first try is using generated function to run instruction.

PR #141, opened 2022-05-07
Update args at copy

FredericWantiez:
In `AdvancedPS` we often need to mutate the input of the the `TapedFunction` to handle random streams. It would be great if we could do something like:
```julia
function (model::Model) (rng::Random.AbstractRNG)
   ...
end

rng1 = Random.MersenneTwister()
task = Litbtask.TapedTask(model, rng1)

Libtask.consume(task)

new_rng = Random.MersenneTwister()
task2 = Libtask.copy(task, new_rng)

Libtask.consume(task) # Generates random numbers from rng1
Libtask.consume(task2) # Generates random numbers from new_rng
```

but for this to work we need to update the `args...` and ` bindings` appropriately when we `copy` the running task. 

This is just a draft, I'm not sure about all the implications of this change, especially with #138

### DIFF
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -373,9 +373,16 @@ function copy_bindings(old::Dict{Symbol, Any})
     return newb
 end
 
-function Base.copy(tf::TapedFunction)
+function Base.copy(tf::TapedFunction, args...)
     # create a new uninitialized TapedFunction
     new_tf = TapedFunction(tf)
     new_tf.bindings = copy_bindings(tf.bindings)
+    if !isempty(args)
+        length(args) > tf.arity && throw("Number of updates for `args` is greater than generated function arity $(tf.arity)")
+        for (i, arg) in enumerate(args)
+            slot = Symbol("_", i + 1)
+            haskey(tf.bindings, slot) && _update_var!(tf, slot, arg)
+       end
+    end
     return new_tf
 end
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -373,16 +373,9 @@ function copy_bindings(old::Dict{Symbol, Any})
     return newb
 end
 
-function Base.copy(tf::TapedFunction, args...)
+function Base.copy(tf::TapedFunction)
     # create a new uninitialized TapedFunction
     new_tf = TapedFunction(tf)
     new_tf.bindings = copy_bindings(tf.bindings)
-    if !isempty(args)
-        length(args) > tf.arity && throw("Number of updates for `args` is greater than generated function arity $(tf.arity)")
-        for (i, arg) in enumerate(args)
-            slot = Symbol("_", i + 1)
-            haskey(tf.bindings, slot) && _update_var!(tf, slot, arg)
-       end
-    end
     return new_tf
 end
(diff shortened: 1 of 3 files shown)


src/tapedfunction.jl
```diff
@@ -373,9 +373,16 @@ function copy_bindings(old::Dict{Symbol, Any})
     return newb
 end
 
-function Base.copy(tf::TapedFunction)
+function Base.copy(tf::TapedFunction, args...)
     # create a new uninitialized TapedFunction
     new_tf = TapedFunction(tf)
     new_tf.bindings = copy_bindings(tf.bindings)
+    if !isempty(args)
```
yebai, line 380: It looks correct in terms of functionality. But the interface is quite general, i.e. we can pass in any `args...` here. We could risk overriding variables in `bindings` with incorrect values silently. Maybe we can think of some assertation to check whether arguments are matching?

KDr2 (commented):
I think storing the args in `TapedTask` will be more intuitive and it can keep the API unchanged.

```diff
diff --git a/src/tapedtask.jl b/src/tapedtask.jl
index 71656a3..8b10506 100644
--- a/src/tapedtask.jl
+++ b/src/tapedtask.jl
@@ -6,6 +6,7 @@ end
 struct TapedTask{F}
     task::Task
     tf::TapedFunction{F}
+    args
     produce_ch::Channel{Any}
     consume_ch::Channel{Int}
     produced_val::Vector{Any}
@@ -13,10 +14,11 @@ struct TapedTask{F}
     function TapedTask(
         t::Task,
         tf::TapedFunction{F},
+        args,
         produce_ch::Channel{Any},
         consume_ch::Channel{Int}
     ) where {F}
-        new{F}(t, tf, produce_ch, consume_ch, Any[])
+        new{F}(t, tf, args, produce_ch, consume_ch, Any[])
     end
 end

@@ -161,7 +163,7 @@ Base.IteratorEltype(::Type{<:TapedTask}) = Base.EltypeUnknown()

 function Base.copy(t::TapedTask)
     tf = copy(t.tf)
-    new_t = TapedTask(tf)
+    args = do_tape_copy_on(t.args...) # !!!
+    new_t = TapedTask(tf, args...) 
     storage = t.task.storage::IdDict{Any,Any}
     new_t.task.storage = copy(storage)
     new_t.task.storage[:tapedtask] = new_t
```

src/tapedfunction.jl
KDr2: This has another issue: we make it the user's responsibility to ensure that the `args` are correctly dealt with (if they should be copied or not), but not dealt with by `copy(::TapedTask)` automaticallly.

src/tapedtask.jl
```diff
@@ -159,9 +161,15 @@ Base.IteratorEltype(::Type{<:TapedTask}) = Base.EltypeUnknown()
...
+function Base.copy(t::TapedTask; args=())
     tf = copy(t.tf)
-    new_t = TapedTask(tf)
+    task_args = if length(args) > 0
+        typeof(args) == typeof(t.args) || error("bad arguments")
+        args
+    else
+        tape_copy.(t.args)
+    end
+    new_t = TapedTask(tf, real_args...)
```
yebai, line 172:
```suggestion
    new_t = TapedTask(tf, task_args...)
```

yebai (approved):
Many thanks, @KDr2 and @FredericWantiez- it looks good and I'm happy to merge once CI passes.

PR #143, opened 2022-05-11
Forbid to copy started task with new args

### LINKED ISSUE #142: Copy args
Seems like #141 has a small issue with copying the `args` around. From the last update:
```julia
function f(x)
  for i in 1:3
    produce(i + x)
  end
end

ttask = TapedTask(f, 1)
consume(ttask) # 2

ttask2 = copy(ttask)
consume(ttask2) # 3

ttask3 = copy(ttask; args=(5,))
consume(ttask3) # 3 (!!) x is still 1
```
but this runs fine
```julia
function f(x)
  for i in 1:3
    produce(i + x)
  end
end

ttask = TapedTask(f, 1)

ttask2 = copy(ttask)
consume(ttask2) # 2

ttask3 = copy(ttask; args=(5,))
consume(ttask3) # 6 correct update
```

KDr2:
Fix #142

### DIFF
--- a/src/tapedtask.jl
+++ b/src/tapedtask.jl
@@ -162,6 +162,8 @@ Base.IteratorEltype(::Type{<:TapedTask}) = Base.EltypeUnknown()
 # copy the task
 
 function Base.copy(t::TapedTask; args=())
+    length(args) > 0 && t.tf.counter >1 &&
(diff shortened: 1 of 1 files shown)


yebai:
thanks, @KDr2!

PR #144, opened 2022-05-14
Fix #142

### LINKED ISSUE #142: Copy args

KDr2:
Fix #142

### DIFF
--- a/src/tapedtask.jl
+++ b/src/tapedtask.jl
@@ -166,10 +166,14 @@ function Base.copy(t::TapedTask; args=())
         error("can't copy started task with new arguments")
     tf = copy(t.tf)
     task_args = if length(args) > 0
+        # this cond implies t.tf.counter == 0, i.e., the task is not started yet
         typeof(args) == typeof(t.args) || error("bad arguments")
         args
     else
-        tape_copy.(t.args)
+        # the task is running, we just need a args placeholder
+        # it would be better (but not necessary?) to find the
+        # real args from the copied bindings
+        t.args
     end
     new_t = TapedTask(tf, task_args...)
     storage = t.task.storage::IdDict{Any,Any}
--- a/src/tapedtask.jl
+++ b/src/tapedtask.jl
@@ -173,8 +173,9 @@ function Base.copy(t::TapedTask; args=())
         if t.tf.counter > 1
             # the task is running, we find the
             # real args from the copied bindings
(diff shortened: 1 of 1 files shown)


src/tapedtask.jl
```diff
@@ -166,10 +166,20 @@ function Base.copy(t::TapedTask; args=())
...
+        # this cond implies t.tf.counter == 0, i.e., the task is not started yet
         typeof(args) == typeof(t.args) || error("bad arguments")
         args
     else
-        tape_copy.(t.args)
+        if t.tf.counter > 1
+            # the task is running, we find the
+            # real args from the copied bindings
+            map(1:length(t.args)) do i
+                get(tf.bindings, Symbol("_", i + 1), t.args[i])
```
FredericWantiez, line 177: Why would we have args in t.args but not in the bindings ?

KDr2: We have all local variables in the bindings, including arguments which are used. (If an argument is not used in the function, it will not be in the bindings, like `x` in `f(x) = 1`).

KDr2: In the bindings, arguments are indicated with keys that have the pattern `:_<N>`.

src/tapedtask.jl
```diff
@@ -166,10 +166,20 @@ function Base.copy(t::TapedTask; args=())
...
+            end
+        else
+            # the task is not started yet, but no args is given
+            tape_copy.(t.args)
```
FredericWantiez, line 181: `tf.bindings` is not populated if `tf.counter = 1` ? Otherwise we're calling `tape_copy` twice

KDr2: Yes, if `tf.counter = 1`, the task is not started, so values of its args are not in bindings.

Issue #145, opened 2022-05-19
Track AdvancedPS master branch in Integration Test


FredericWantiez:
We probably need to track the `master` branch of AdvancedPS as well. Looks like #144 is causing issue downstream [failing tests](https://github.com/TuringLang/AdvancedPS.jl/pull/54/checks). Not entirely sure what the error is since the tests are running fine on my local setup.  
https://github.com/TuringLang/Libtask.jl/blob/330fb8210bff8f42f12664af84566af496ebbf9b/.github/workflows/IntegrationTest.yml#L16-L18

yebai:
It looks like we are calling a removed function @KDr2 - could you take a look, please?

FredericWantiez:
Looks like it's a `CodeInfoTools` error when running tests with `coverage=true`. @KDr2 you found the issue earlier here I think [issue](https://github.com/JuliaCompilerPlugins/CodeInfoTools.jl/issues/18)

This task
```julia
function (m::Model)(rng)
    m.a = rand(rng, Normal())
    AdvancedPS.observe(Normal(), m.a)

    m.b = rand(rng, Normal())
    return AdvancedPS.observe(Normal(), m.b)
end
```
produces the right type annotation when running `using Pkg; Pkg.test()`:

```bash
CodeInfo(
    @ /home/frederic/Documents/projects/turing/AdvancedPS.jl/test/smc.jl:163 within `Model`
1 ─ %1  = Main.Normal()::Core.Const(Normal{Float64}(μ=0.0, σ=1.0))
│   %2  = Main.rand(rng, %1)::Float64
│         Base.setproperty!(m, :a, %2)::Any
│   @ /home/frederic/Documents/projects/turing/AdvancedPS.jl/test/smc.jl:164 within `Model`
│   %4  = AdvancedPS.observe::Core.Const(AdvancedPS.observe)
│   %5  = Main.Normal()::Core.Const(Normal{Float64}(μ=0.0, σ=1.0))
│   %6  = Base.getproperty(m, :a)::Float64
│         (%4)(%5, %6)::Any
│   @ /home/frederic/Documents/projects/turing/AdvancedPS.jl/test/smc.jl:166 within `Model`
│   %8  = Main.Normal()::Core.Const(Normal{Float64}(μ=0.0, σ=1.0))
│   %9  = Main.rand(rng, %8)::Float64
│         Base.setproperty!(m, :b, %9)::Any
[...]

PR #146, opened 2022-05-23
workaround for code coverage

KDr2:
Try to fix https://github.com/TuringLang/Libtask.jl/issues/145 and https://github.com/JuliaCompilerPlugins/CodeInfoTools.jl/issues/18.

### DIFF
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -3,6 +3,13 @@
 abstract type AbstractInstruction end
 const RawTape = Vector{AbstractInstruction}
 
+function _infer(f, args_type)
+    ir0 = code_typed(f, Tuple{args_type...}, optimize=false)[1][1]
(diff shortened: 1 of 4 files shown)


src/tapedfunction.jl
```diff
@@ -3,6 +3,13 @@
 abstract type AbstractInstruction end
 const RawTape = Vector{AbstractInstruction}
 
+function _infer(f, args_type)
+    ir0 = code_typed(f, Tuple{args_type...}, optimize=false)[1][1]
```
yebai, line 7: The `..[1][1]` part is not very readable. Maybe add a comment, or change it into `first()` calls?

yebai (approved):
Thanks, @KDr2 - it looks good overall.

src/tapedfunction.jl
KDr2: updated.

PR #156, opened 2022-08-18
Translate new construction of NamedTuple

KDr2:
Fix https://github.com/TuringLang/Turing.jl/issues/1873

### DIFF
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -369,6 +369,15 @@ function translate!!(var::IRVar, line::Core.SlotNumber,
     return Instruction(func, input, output)
 end
 
+function translate!!(var::IRVar, line::NTuple{N, Symbol},
+                     bindings::Bindings, isconst::Bool, ir) where {N}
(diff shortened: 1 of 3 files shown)


src/tapedfunction.jl
```diff
@@ -437,7 +446,7 @@ function translate!!(var::IRVar, line::Expr,
     end
 end
 
-function translate!!(var, line, bindings, ir)
+function translate!!(var, line, bindings, i, ir)
```
yebai, line 449: What's this additional argument `i'?

yebai (approved):
Thanks, @KDr2 - it looks good other than the minor clarification question below.

src/tapedfunction.jl
KDr2, line 449:
```suggestion
function translate!!(var, line, bindings, isconst, ir)
```

KDr2: the signature of `translate!!` changed in a past PR, I forgot to update this method then.

PR #157, opened 2022-10-30
Union types

FredericWantiez:
Not sure it's an issue but seems a bit weird to override the `deepcopy_types` from what the user provides. 
On `master` this produces the wrong result because `Ref` and `Array` are not deepcopied anymore
```julia
using Libtask

struct DummyType end

function f(start::Int)
    t = [start]
    while true
        produce(t[1])
        t[1] = 1 + t[1]
    end
end

ttask = TapedTask(f, 0; deepcopy_types=DummyType)
println(consume(ttask)) # 0 Correct 

ttask2 = copy(ttask)
println(consume(ttask2)) # 1 Correct

println(consume(ttask)) # 2 Incorrect
println(consume(ttask2)) # 3 Incorrect
```

Maybe we can assume that `Array` and `Ref` would always be deepcopied unless explicitly overloaded in `tape_deepcopy` ?

### DIFF
--- a/test/tape_copy.jl
+++ b/test/tape_copy.jl
@@ -172,7 +172,7 @@
         @test y[][2] == 19
     end
 
-    @testset "override deepcopy_types" begin
+    @testset "override deepcopy_types #57" begin
         struct DummyType end
 
         function f(start::Int)
--- a/test/tape_copy.jl
+++ b/test/tape_copy.jl
@@ -171,4 +171,25 @@
         y[][2] = 19
         @test y[][2] == 19
     end
+
+    @testset "override deepcopy_types" begin
+        struct DummyType end
+
+        function f(start::Int)
+            t = [start]
+            while true
+                produce(t[1])
+                t[1] = 1 + t[1]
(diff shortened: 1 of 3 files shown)


test/tape_copy.jl
```diff
@@ -171,4 +171,25 @@
         y[][2] = 19
         @test y[][2] == 19
     end
+
+    @testset "override deepcopy_types" begin
```
yebai, line 175:
```suggestion
    @testset "override deepcopy_types #57" begin
```

yebai:
I am happy with the proposed change. 

Just to clarify, it is not possible to override the copying behaviour for `Array` and `Ref` without modifying the global variable `BASE_COPY_TYPES`? Is that intended?

yebai (approved):

FredericWantiez:
Yes, not sure it's the best way but I don't really see a case where we want shallow copied arrays, and we can alsways overload `tape_deepcopy(x::Array)` explicitly. Happy to change

PR #159, opened 2022-11-02
Fix for #158

### LINKED ISSUE #158: Cache key in tapedfunction
Was looking at
https://github.com/TuringLang/Libtask.jl/blob/5ec9f2bafcb8f2716741efb2845efda5a465b3fd/src/tapedfunction.jl#L65

Shouldn't it be something like this instead ?
```julia
cache_key = (f, deepcopy_types, args_type...) 
```
Otherwise this is giving counterintuitive results:
```julia
struct Dummy 
    a::Array
end

function f(d::Dummy)
    produce(d.a)
end

d = Dummy([1])
t = TapedTask(f, d) # No Dummy in deepcopy_types
t2 = TapedTask(f, d; deepcopy_types=Dummy) # Dummy won't be part of the deepcopy_types
```

FredericWantiez:
Would close #158

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 license = "MIT"
 desc = "Tape based task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
-version = "0.8.3"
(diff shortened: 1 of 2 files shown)


yebai (approved):

yebai:
thanks @FredericWantiez!

PR #162, opened 2023-04-04
Fix #161

### LINKED ISSUE #161: `TypeError` in `tape_copy` for newer versions of Julia
Seems like `allocate_binding!` fails on a `TypeError` with newer versions of Julia. For example, on julia `1.9`, the `tape_copy` tests fail on something similar to this:
```julia
  Got exception outside of a @test
  TypeError: in Type, in parameter, expected Type, got a value of type Pair{String, Int64}
  Stacktrace:
    [1] allocate_binding!(var::Core.SSAValue, bindings::Vector{Any}, c::Core.Const)
      @ Libtask ~/Documents/projects/turing/Libtask.jl/src/tapedfunction.jl:299
    [2] bind_var!(var::Core.SSAValue, bindings::Vector{Any}, ir::Core.CodeInfo)
```

KDr2:
Fix #161.

### DIFF
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -297,7 +297,8 @@ function bind_var!(var::Core.SSAValue, bindings::Bindings, ir::Core.CodeInfo)
 end
 
 allocate_binding!(var, bindings::Bindings, c::Core.Const) =
-    allocate_binding!(var, bindings, _loose_type(Type{c.val}))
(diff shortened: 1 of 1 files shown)


KDr2:
Fixed on Julia 1.9, but `Core.TypedSlot` is removed on Julia Version 1.10.0-DEV.944 :(

PR #164, opened 2023-06-21
Prototype implementation of recursion

willtebbutt:
A basic implementation of recursion against which tweaks to Turing's submodel interface can be tested.

TODO:
- [ ] ~~add context variable to enable different primitives in different situations~~
- [x] improve documentation for `is_primitive`
- [x] bump version
- [x] implement cache for `TapedFunction` based on `func` and argument types to avoid re-tracing each time

Maybe TODO:
- [ ] find a better way of testing

@torfjelde when you get a minute, could you please check to see whether this does what you need it to do?

PR #165, opened 2023-06-22
[WIP] A way to add kwarg support to #164

torfjelde:
I had a branch locally with some initial work on kwarg-support, and now, thanks to #164, it works!

``` julia
julia> f(x; y=2) = (produce(2x + y); return 2x)

julia> g(x) = f(f(x))

julia> Libtask.is_primitive(::typeof(f), x) = false

julia> task = Libtask.TapedTask(g, 1)
TapedTask{typeof(g), Tuple{Int64}}(Task (runnable) @0x00007f9f16a160e0, TapedFunction:
* .func => g
* .ir   =>
------------------
CodeInfo(
1 ─ %1 = Main.f(x)::Int64
│   %2 = Main.f(%1)::Int64
└──      return %2
)
------------------
, (1,), Channel{Any}(0), Channel{Int64}(0), Any[])

julia> consume(task)
4

julia> consume(task)
6

julia> consume(task)
```

A few _very_ important drawbacks:
- [ ] Currently only work on Julia 1.9, but it should be possible to make it work on previous versions by making use of `Core.kwftype`.
- [ ] Will have name-clashes with `cache` and `deepcopy_types` that are used in Libtask.jl. This can be resolved by a) renaming those to something less common, e.g. `__libtask__cache__` or something, and b) throwing an error if we hit it.

EDIT: I wanted to make this PR to your branch @willtebbutt but couldn't because that branch is on your fork.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 license = "MIT"
 desc = "Tape based task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
-version = "0.8.6"
+version = "0.8.7"
 
 [deps]
 FunctionWrappers = "069b7b12-0de2-55c6-9aab-29f3d0a68a2e"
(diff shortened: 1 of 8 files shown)


torfjelde:
TBH I think I prefer this to making changes in DynamicPPL, as originally discussed. This way we keep all the complexity in Libtask, rather than shifting some of the complexity of Libtask into the design of DynamicPPL.jl

yebai:
@willtebbutt, does #179 fix the support for kwargs?

willtebbutt:
I'm unsure whether it currently does -- it shouldn't be too hard to add though.

yebai:
It would be good to support it.

PR #171, opened 2024-01-17
Subtape + Fix Recursion

markus7800:
Hello,

I am not too familiar with the code base, but I updated the subtaping functionality to work for my own project.
Maybe you find these fixes useful.

- Added `subtapes` property to `TapedFunction`
This is important for task copying to work. We also have to copy the subtapes to continue in the correct instruction in the copied task (e.g. if we want to continue in a subtape).
- Remove translation optimisation based on constant return types of calls. A call may contain a produce statement but returns a constant. In the old implementation this call would have been optimised.
- Added a few test cases for subtapes that did not pass in the old implementation, but now do.

Best,
Markus

### DIFF
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -529,5 +529,5 @@ end
 # but if we instantiate new TapedTask, we have to reset the counters of cached tapes
 function reset_counters!(tf::TapedFunction)
     tf.counter = 1
-    reset_counters!.(values(tf.subtapes))
+    foreach(reset_counters!, values(tf.subtapes))
 end
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -420,12 +420,6 @@ function translate!!(var::IRVar, line::Expr,
         # inferred as constants: we only optimize primitive and datatype constants for now. For
         # optimised function calls, we will evaluate the function at compile-time and cache results.
 
-        # even if return value is const. The call may has side-effects.
-        # In particular, it may contain produce statements.
-        # if isconst
-        #     v = ir.ssavaluetypes[var.id].val
-        #     _canbeoptimized(v) && return _const_instruction(var, v, bindings, ir)
-        # end
         args = map(_bind_fn, line.args)
         # args[1] is the function
         func = line.args[1]
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -523,3 +523,10 @@ function Base.copy(tf::TapedFunction)
     new_tf.subtapes = IdDict{Any,TapedFunction}(func => copy(subtape) for (func, subtape) in tf.subtapes)
     return new_tf
 end
+
+# when copying we want to keep the counters
+# but if we instantiate new TapedTask, we have to reset the counters of cached tapes
+function reset_counters!(tf::TapedFunction)
+    tf.counter = 1
+    reset_counters!.(values(tf.subtapes))
+end
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -222,7 +222,6 @@ function (instr::Instruction{F})(tf::TapedFunction, callback=nothing) where F
         else
             tf_inner = get!(tf.subtapes, instr, TapedFunction(func, inputs..., cache=true))
             tf_inner(inputs...; callback=callback, continuation=true)
-            @assert tf_inner.retval_binding_slot != 0 # tf_inner should return
         end
         _update_var!(tf, instr.output, output)
         tf.counter += 1
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -67,7 +67,8 @@ mutable struct TapedFunction{F, TapeType}
         if cache && haskey(TRCache, cache_key) # use cache
             cached_tf = TRCache[cache_key]::TapedFunction{F, T}
             tf = copy(cached_tf)
-            tf.counter = 1
+            # we have to reset the counters of cached tapes (also the counters of subtapes)
+            reset_counters!(tf)
             return tf
         end
         ir = _infer(f, args_type)
--- a/src/tapedfunction.jl
+++ b/src/tapedfunction.jl
@@ -58,6 +58,7 @@ mutable struct TapedFunction{F, TapeType}
     arg_binding_slots::Vector{Int} # arg indices in binding_values
     retval_binding_slot::Int # 0 indicates the function has not returned
     deepcopy_types::Type # use a Union type for multiple types
+    subtapes::IdDict{Any,TapedFunction}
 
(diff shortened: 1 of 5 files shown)


src/tapedfunction.jl
```diff
@@ -219,8 +220,9 @@ function (instr::Instruction{F})(tf::TapedFunction, callback=nothing) where F
         output = if is_primitive(func, inputs...)
             func(inputs...)
         else
-            tf_inner = TapedFunction(func, inputs..., cache=true)
-            tf_inner(inputs...; callback=callback)
+            tf_inner = get!(tf.subtapes, instr, TapedFunction(func, inputs..., cache=true))
+            # continuation=false breaks "Multiple func calls subtapes" and "Copying task with subtapes"
+            tf_inner(inputs...; callback=callback, continuation=true)
```
torfjelde, line 225: IIUC `continuation=true` is now needed because we might be calling the same `TapedFunction` multiple times?

EDIT: In contrast to before when were just constructing a new one every time.

src/tapedfunction.jl
```diff
@@ -416,10 +418,13 @@ function translate!!(var::IRVar, line::Expr,
...
-            v = ir.ssavaluetypes[var.id].val
-            _canbeoptimized(v) && return _const_instruction(var, v, bindings, ir)
-        end
+
+        # even if return value is const. The call may has side-effects.
+        # In particular, it may contain produce statements.
+        # if isconst
+        #     v = ir.ssavaluetypes[var.id].val
+        #     _canbeoptimized(v) && return _const_instruction(var, v, bindings, ir)
+        # end
```
torfjelde, line 427:
```suggestion
```

Since you've added tests, I think it's fine to just remove this entirely; if someone tries to add it back for some reason without anyone advising against it, tests will fail :+1:

test/tapedtask.jl
```diff
@@ -155,5 +155,59 @@
                 @test ttask2.task.exception isa BoundsError
             end
         end
+
+        @testset "Too much producers" begin
```
torfjelde, line 159:
```suggestion
        @testset "Too many producers" begin
```

test/tf.jl
```diff
@@ -37,9 +37,75 @@ Libtask.is_primitive(::typeof(foo), args...) = false
...
     @testset "recurse into function" begin
-        tf = Libtask.TapedFunction(bar, 5.0)
-        count = 0
-        tf(4.0; callback=() -> (count += 1))
-        @test count == 9
+        # tf = Libtask.TapedFunction(bar, 5.0)
+        # count = 0
+        # tf(4.0; callback=() -> (count += 1))
+        # @test count == 9
+
```
torfjelde, line 44:
```suggestion
```

src/tapedfunction.jl
```diff
@@ -515,5 +520,13 @@ end
...
     new_tf.binding_values = copy_bindings(tf.binding_values, tf.deepcopy_types)
+    new_tf.subtapes = IdDict{Any,TapedFunction}(func => copy(subtape) for (func, subtape) in tf.subtapes)
     return new_tf
 end
+
+# when copying we want to keep the counters
+# but if we instantiate new TapedTask, we have to reset the counters of cached tapes
+function reset_counters!(tf::TapedFunction)
+    tf.counter = 1
+    reset_counters!.(values(tf.subtapes))
```
torfjelde, line 531: Maybe
```suggestion
    foreach(reset_counters!, values(tf.subtapes))
```
?

torfjelde (commented):
Hey Markus!

First of all, thank you for helping out with this!

Because my familiarity with Libtask is pretty bad, I'm just leaving some comments + trying to understand what's going on here; I'll leave the approval to someone else:)

Am I understanding it correctly that we need to keep the subtapes around because of scenarios such as the following. Suppose we have

``` julia
function f()
    produce(1)
    g()
    produce(2)
end

function g()
    produce(3)
    produce(4)
    produce(5)
end

Libtask.is_primitive(::typeof(g), args...) = false
```

and we decide to make a `copy` after hitting `produce(4)`. If we then try to execute the instruction for `g` in `f` again, we'll end up constructing an entirely new instance of instruction `g` (in the current version, not yours) and so will start from `produce(4)` again rather than `produce(5)`? 

EDIT: Simpler example (on `#master`)

```julia
julia> f() = g()

julia> function g()
           produce(1)
           produce(2)
       end

julia> ttask = TapedTask(f);

julia> iterate(ttask)
(2, nothing)

julia> Libtask.is_primitive(::typeof(g), args...) = false

julia> ttask = TapedTask(f);

julia> iterate(ttask)
(1, nothing)

julia> iterate(ttask)
(2, nothing)

julia> ttask2 = copy(ttask);

julia> iterate(ttask2)  #  BAD: starts from the beginning of `g` 
(1, nothing)
```

src/tapedfunction.jl
markus7800: I think your understanding is correct.

In the tape, a function call constitutes a single instruction.
Previous to my changes, this function call was only allowed to call `produce(val)` once.
I think the reason was that without subtapes, we were only able to continue the execution *after* the call instruction.
This was safe to do if we only allow one `produce` and throw an error otherwise.

By flagging a function `g` as non-primitive `is_primitive(typeof(g),...) = false`, we tell Libtask that we want to be able to interrupt the execution in `g` (e.g. at `produce(4)`).

For this to work any parent tape function `f` (the caller) has to actually own the subtape of `g`. 
The instruction `counter` of `g` has to be preserved, such that with `continuation=true` we continue exactly at the correct instruction *in* `g`. And yes, `g` is called multiple times with an updated `counter`.

Thus, when we fork/copy a task, we also have to copy all subtapes with their counters.

Also, when reusing a cached tape we have to reset all the counters of the subtapes `counter = 1`

src/tapedfunction.jl
```diff
@@ -219,8 +221,9 @@ function (instr::Instruction{F})(tf::TapedFunction, callback=nothing) where F
         output = if is_primitive(func, inputs...)
             func(inputs...)
         else
-            tf_inner = TapedFunction(func, inputs..., cache=true)
-            tf_inner(inputs...; callback=callback)
+            tf_inner = get!(tf.subtapes, instr, TapedFunction(func, inputs..., cache=true))
```
devmotion, line 224: This will always construct a `TapedFunction` object, regardless of whether `tf.subtapes[instr]` exists or not. Maybe instead we could use
```suggestion
            tf_inner = get!(tf.subtapes, instr) do
                TapedFunction(func, inputs...; cache=true)
            end
```

src/tapedfunction.jl
torfjelde: That makes sense; thank you! And again, great work:)

Issue #174, opened 2024-10-21
Unknown expression / IR code `typeof(var) = Core.SSAValue`


penelopeysm:
This code fails on both 1.10 and 1.11:

```julia
using Turing
@model function g()
    a ~ Normal(0, 1)
    b = 2
end
sample(g(), Gibbs(PG(10, :a), HMC(0.01, 4, :b)), 10)
```

Note that if we change `b` to `~ Normal(0, 1)` as well, the code runs fine on both 1.10 and 1.11, so the bug is something to do with the assignment in the model.

In both cases the error is similar

```
┌ Error: Unknown IR code:
│   typeof(var) = Core.SSAValue
│   var = :(%85)
│   typeof(line) = Int64
│   line = 2
└ @ Libtask ~/.julia/packages/Libtask/0vMZ5/src/tapedfunction.jl:449
Sampling 100%|████████████████████████████████████████████████████████████████████| Time: 0:00:02
ERROR: Unknown IR code
Stacktrace:
  [1] translate!!(var::Core.SSAValue, line::Int64, bindings::Vector{Any}, isconst::Bool, ir::Core.CodeInfo)
    @ Libtask ~/.julia/packages/Libtask/0vMZ5/src/tapedfunction.jl:450
  [2] translate!(tape::Vector{Libtask.AbstractInstruction}, ir::Core.CodeInfo)
    @ Libtask ~/.julia/packages/Libtask/0vMZ5/src/tapedfunction.jl:320
  [3] Libtask.TapedFunction{…}(::typeof(g), ::DynamicPPL.Model{…}, ::Vararg{…}; cache::Bool, deepcopy_types::Type)
    @ Libtask ~/.julia/packages/Libtask/0vMZ5/src/tapedfunction.jl:73
  [4] Libtask.TapedFunction(::Function, ::DynamicPPL.Model{…}, ::Vararg{…}; cache::Bool, deepcopy_types::Type)
    @ Libtask ~/.julia/packages/Libtask/0vMZ5/src/tapedfunction.jl:80
  [5] TapedTask(::Function, ::DynamicPPL.Model{…}, ::Vararg{…}; deepcopy_types::Type)
    @ Libtask ~/.julia/packages/Libtask/0vMZ5/src/tapedtask.jl:76
  [6] TapedTask(::Turing.Essential.TracedModel{…}, ::AdvancedPS.TracedRNG{…}; kwargs::@Kwargs{…})
[...]

PR #175, opened 2024-10-22
static_parameter expr

KDr2:
Fix https://github.com/TuringLang/Libtask.jl/issues/174

### DIFF
--- a/.github/workflows/Testing.yaml
+++ b/.github/workflows/Testing.yaml
@@ -13,6 +13,7 @@ jobs:
         version:
           - '1.7'
           - '1'
+          - '1.11.1'
           - 'nightly'
         os:
           - ubuntu-latest
--- a/.github/workflows/Testing.yaml
+++ b/.github/workflows/Testing.yaml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.10'
+          - '1.7'
           - '1'
           - 'nightly'
         os:
--- a/.github/workflows/Testing.yaml
+++ b/.github/workflows/Testing.yaml
@@ -13,7 +13,6 @@ jobs:
         version:
           - '1.7'
           - '1'
-          - '1.11.1'
           - 'nightly'
         os:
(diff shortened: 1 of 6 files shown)


willtebbutt (commented):
This LGTM. You'll need to bump the version number before merging though so that we can tag a release.

penelopeysm:
Yes, it does fix the examples in #174! To be precise, the first one still gives an error, but it now gives another error that isn't due to Libtape (it's a meaningless model after all).

penelopeysm:
Haha, sorry, I just edited the comment (probably after you saw it) to say that I hadn't worked that one out yet. I'm looking at it now!

Thank you for looking into this so quickly!!

penelopeysm:
@KDr2 I think it's the `::Type{T}` that makes this fail on master (but is fixed on this PR):

```julia
using Libtask
function g(::Type{T}) where {T}
    a = zeros(T, 10)
end
tf = Libtask.TapedFunction(g, Float64)
```

penelopeysm:
```julia
using Libtask
function g(::Type{T}) where {T}
    a = zeros(T, 10)
end
tf = Libtask.TapedFunction(g, Float64)
```

The above seems fine, but then subsequently calling `r = tf(Float64)` fails, at least on <=1.10. This isn't a regression introduced by this PR, it's also on master.

KDr2:
> This isn't a regression introduced by this PR, it's also on master.

Fixed, but what happened to the Integeration test? It passes now...

penelopeysm:
It's probably because of Hong's commit which bumped the minor version, and Turing.jl's compat entry for Libtask doesn't allow it to be tested with this.

I'm curious as to whether we can keep 1.7 compat here, if you don't mind, I'll test it out :)

KDr2:
> we can keep 1.7 compat

We can keep both 1.7 and 1.10 in the CI unit tests.

penelopeysm:
Thank you again to all of you who contributed here ❤️

Issue #186, opened 2025-06-06
DynamicCallable issue with varargs


mhauru:
MWE:
```julia
julia> function g()
                  if randn() > 0
                      f = Base.vect
                  else
                      f = tuple
                  end
                  produce(f(0.1, 0.2))
              end
g (generic function with 1 method)

julia> show(Libtask.consume(Libtask.TapedTask(nothing, g)))
ERROR: MethodError: no method matching (::Core.OpaqueClosure{Tuple{typeof(Base.vect), Tuple{…}}, Vector{Float64}})(::typeof(Base.vect), ::Float64, ::Float64)
This error has been manually thrown, explicitly, so the method may exist but be intentionally marked as unimplemented.

Closest candidates are:
  (::Core.OpaqueClosure{Tuple{typeof(Base.vect), Tuple{Float64, Float64}}, Vector{Float64}})(::typeof(Base.vect), ::Tuple{Float64, Float64}) (method too new to be called from this world context.)
   @ Core :0

Stacktrace:
 [1] (::MistyClosures.MistyClosure{Core.OpaqueClosure{Tuple{…}, Vector{…}}})(::Function, ::Float64, ::Float64)
   @ MistyClosures ~/.julia/packages/MistyClosures/4rFSA/src/MistyClosures.jl:22
 [2] (::Libtask.DynamicCallable{Dict{Any, Any}})(::Function, ::Float64, ::Float64)
   @ Libtask ~/projects/Libtask.jl/src/copyable_task.jl:1144
 [3] g
   @ ./REPL[47]:2 [inlined]
 [4] (::Tuple{…})(none::typeof(g))
   @ Base.Experimental ./<missing>:0
 [5] consume(t::TapedTask{Nothing, Tuple{typeof(g)}, MistyClosures.MistyClosure{Core.OpaqueClosure{Tuple{typeof(g)}, Any}}})
   @ Libtask ~/projects/Libtask.jl/src/copyable_task.jl:338
 [6] top-level scope
   @ REPL[48]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

[...]

PR #187, opened 2025-06-13
Issue #185

### LINKED ISSUE #185: `produce` not being hit with type unstable variable
MWE:
```julia
julia> function g()
                  if randn() > 0
                      x = 2
                  else
                      x = 0.1
                  end
                  produce(x)
              end

julia> show(Libtask.consume(Libtask.TapedTask(nothing, g)))
nothing
```

The fact that the type of `x` depends on runtime values seems to be significant. The following variants work as expected:
```julia
julia> function g()
                  if randn() > 0
                      x = 0.2
                  else
                      x = 0.1
                  end
                  produce(x)
              end

julia> show(Libtask.consume(Libtask.TapedTask(nothing, g)))
0.2
```
and
```julia
julia> function g()
                  if randn() isa Float64
                      x = 1
                  else
                      x = 0.1
                  end
                  produce(x)
              end

julia> show(Libtask.consume(Libtask.TapedTask(nothing, g)))
1
```

mhauru:
This fixes #185, and adds a helpful error message if one tries to call `TapedTask(_, produce, x)` (which I suspect we can't make work for somewhat deep reasons).

The problem in #185 is that sometimes the IR has `:invoke` expressions where the thing being invoked isn't literally `produce`, but rather a IR/BB ID referencing a variable that is `produce`. In these cases `is_produce_stmt` fails to recognise that this is indeed a call to produce. At least in the cases that my MWE brings up, however, the `:invoke` still carries with it the knowledge that this is `MethodInstance` of `produce`, so I changed the check to look for that.

I don't know what I'm doing when I'm messing with things like `MethodInstance`s, so would be very good to get @willtebbutt's eyes on this. Maybe there's a more robust way to fix this?

Closes #185

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 license = "MIT"
 desc = "Tape based task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
-version = "0.9.1"
+version = "0.9.2"
 
 [deps]
 MistyClosures = "dbe65cb8-6be2-42dd-bbc5-4196aaced4f4"
--- a/src/copyable_task.jl
+++ b/src/copyable_task.jl
@@ -72,6 +72,13 @@ the current world age, will make a copy of an existing `MistyClosure`. If not,
 will derive it from scratch (derive the IR + compile it etc).
 """
 function build_callable(sig::Type{<:Tuple})
+    if sig <: Tuple{typeof(produce),Any}
+        msg = """
+            Can not construct a TapedTask for a 'naked' call to `produce`.
+             Please wrap the call to `produce` in a function, and construct a
+             TapedTask from that function."""
+        throw(ArgumentError(msg))
+    end
     key = CacheKey(Base.get_world_counter(), sig)
     if haskey(mc_cache, key)
         return fresh_copy(mc_cache[key])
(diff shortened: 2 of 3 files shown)


sunxd3 (approved):
make sense to me, no comments here

willtebbutt:
Hmmm which version of Julia are you producing this IR from? I see the following on 1.11.5:
```julia
julia> Base.code_ircode(g, (Float64,))
1-element Vector{Any}:
2 1 ─ %1 = Base.lt_float(0.0, _2)::Bool                               │╻╷╷ >
  │   %2 = Base.or_int(%1, false)::Bool                               ││╻   <
  └──      goto #3 if not %2                                          │
  2 ─      goto #4                                                    │
  3 ─      nothing::Nothing                                           │
7 4 ┄ %6 = φ (#2 => 2, #3 => 0.1)::Union{Float64, Int64}              │
  │   %7 = Main.produce::Any                                          │
  │   %8 = (%7)(%6)::Any                                              │
  └──      return %8                                                  │
   => Any
```
and a similar thing on 1.10.9. It does look like you've got two methods of `g` your snippet though, whereas I'm running in a clean Julia session. Maybe that's where the difference comes from?

mhauru:
Oooh, nice catch, I had a long-living environment. It scares me that having an unrelated method defined affects the IR code of this method 😬 . Anyway, I reverted back to using the MWE from the original issue, and checked that without the fix in this PR the test fails in a fresh session.

mhauru:
I've bumped version and merged. I'm trying to fix #186, might bundle the fix together in the same release, unless you prefer otherwise.

PR #188, opened 2025-06-17
Fix varargs calls

### LINKED ISSUE #186: DynamicCallable issue with varargs

mhauru:
Another case where I'm changing things that I only have a superficial understanding of. @willtebbutt, could you please check that `check_varargs` makes sense to you, and feel free to propose a better solution.

All the tests added here fail on main.

Closes #186

### DIFF
--- a/docs/src/internals.md
+++ b/docs/src/internals.md
@@ -9,6 +9,7 @@ Libtask.get_type
 Libtask._typeof
 Libtask.replace_captures
 Libtask.BasicBlockCode
+Libtask.check_varargs
 Libtask.opaque_closure
 Libtask.misty_closure
 Libtask.optimise_ir!
--- a/docs/src/internals.md
+++ b/docs/src/internals.md
@@ -9,7 +9,6 @@ Libtask.get_type
 Libtask._typeof
 Libtask.replace_captures
 Libtask.BasicBlockCode
-Libtask.check_varargs
 Libtask.opaque_closure
 Libtask.misty_closure
 Libtask.optimise_ir!
--- a/src/copyable_task.jl
+++ b/src/copyable_task.jl
@@ -64,31 +64,6 @@ function callable_ret_type(sig, produce_types)
 end
 
 """
-    check_varargs(sig, ir)
-
-For a call signature `sig` and the IR code for it, check whether this is a Varargs call.
-
-The need for this arises because the output of `Base.code_ircode_by_type` does not
-distinguish between varargs and tuples, see https://github.com/JuliaLang/julia/issues/58753. Thus we have to go back to the signature that created the IR to check. There are two cases
-that signal that this is indeed a varargs call:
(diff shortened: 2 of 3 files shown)


mhauru:
[This](https://github.com/JuliaLang/julia/issues/58753#issuecomment-2979954978) great tip from @simeonschaub helped me simplify the implemenation a lot. I feel much less insecure about the new version, but would still be good to get Will's comments on this.

PR #196, opened 2025-07-31
Support for Julia v1.12

mhauru:
Work in progress, though most tests now pass. Current known issues are calling varargs functions and and more allocations than on v1.11.


yebai:
https://github.com/JuliaLang/julia/issues/59222

These tests look slightly weird. They might be violating opaque closure assumptions, e.g. dynamic dispatch of `g(xs...)` depends on inputs to the opaque closure itself. Do we need them here?

PR #205, opened 2025-11-05
Use set_valid_worlds! instead of manual optimisation

mhauru:
Replicating the changes made to Mooncake by in https://github.com/chalk-lab/Mooncake.jl/pull/832.

The benchmarks that pass, run on this branch:
```
benchmarking rosenbrock...
  Run Original Function:  185.542 μs (24 allocations: 6.25 MiB)
  Run TapedTask: #produce=1;   6.784 ms (299145 allocations: 10.83 MiB)
benchmarking ackley...
  Run Original Function:  1.154 ms (0 allocations: 0 bytes)
  Run TapedTask: #produce=100000;   60.412 ms (899584 allocations: 21.36 MiB)
benchmarking matrix_test...
  Run Original Function:  97.417 μs (18 allocations: 576.47 KiB)
  Run TapedTask: #produce=1;   468.625 μs (530 allocations: 594.08 KiB)
benchmarking neural_net...
  Run Original Function:  445.707 ns (8 allocations: 576 bytes)
  Run TapedTask: #produce=1;   12.625 μs (168 allocations: 6.27 KiB)
```
on `main`:
```
benchmarking rosenbrock...
  Run Original Function:  169.583 μs (24 allocations: 6.25 MiB)
  Run TapedTask: #produce=1;   2.062 ms (679 allocations: 6.27 MiB)
benchmarking ackley...
  Run Original Function:  1.154 ms (0 allocations: 0 bytes)
  Run TapedTask: #produce=100000;   56.672 ms (699582 allocations: 18.31 MiB)
benchmarking matrix_test...
  Run Original Function:  98.250 μs (18 allocations: 576.47 KiB)
  Run TapedTask: #produce=1;   294.750 μs (529 allocations: 594.06 KiB)
benchmarking neural_net...
  Run Original Function:  430.065 ns (8 allocations: 576 bytes)
  Run TapedTask: #produce=1;   12.041 μs (168 allocations: 6.27 KiB)
```
and on `main` but skipping the manual optimisation and just calling `misty_closure` directly:
```
benchmarking rosenbrock...
  Run Original Function:  173.083 μs (24 allocations: 6.25 MiB)
  Run TapedTask: #produce=1;   6.428 ms (299145 allocations: 10.83 MiB)
benchmarking ackley...
  Run Original Function:  1.153 ms (0 allocations: 0 bytes)
  Run TapedTask: #produce=100000;   61.659 ms (899584 allocations: 21.36 MiB)
benchmarking matrix_test...
[...]

PR #209, opened 2025-11-07
Fix #207

### LINKED ISSUE #207: Libtask v0.9.9 breaks Union types
This is causing Turing PG/SMC to fail on all models.

```julia
using Libtask

g(i::Int) = begin; produce(1); i; end
g(s::String) = begin; produce(2); s; end
Libtask.@might_produce g

h(i::Int)::Int = i
h(s::String)::Int = length(s)

function f(x)
    i_or_s = if x > 0
        3
    else
        "libtask"
    end
    # These lines are just to make sure that `f` itself is type stable.
    # You can trigger the same error by just returning `g(i_or_s)`
    gres = g(i_or_s)
    return h(gres)
end

t = TapedTask(nothing, f, 1)
consume(t) # returns 1
consume(t) # should return `nothing`, but errors on 0.9.9 with:

#=
ERROR: TypeError: in typeassert, expected Base.RefValue{Union{Int64, String}}, got a value of type Base.RefValue{Int64}
Stacktrace:
 [1] deref_phi
   @ ~/ppl/Libtask.jl/src/copyable_task.jl:1229 [inlined]
 [2] f
   @ ./REPL[26]:2 [inlined]
 [3] (::Tuple{…})(_2::typeof(f), _3::Int64)
   @ Base.Experimental ./<missing>:0
 [4] consume(t::TapedTask{Nothing, Tuple{typeof(f), Int64}, MistyClosures.MistyClosure{Core.OpaqueClosure{…}}})
   @ Libtask ~/ppl/Libtask.jl/src/copyable_task.jl:378
 [5] top-level scope
   @ REPL[29]:1
=#
```

It works fine on 0.9.8 so this is caused by #205. You can fix it by moving the call to `gres` inside the if/else, so this works:

```julia
function f(x)
    gres = if x > 0
        g(3)
    else
        g("libtask")
    end
    return h(gres)
end
```

[The immediate Turing breakage is here](https://github.com/TuringLang/Turing.jl/blob/905f7fa79dbcd64497711813c11d4c08136bff33/src/mcmc/particle_mcmc.jl#L410-L424). You can work around it by moving the `tilde_assume!!` call inside the if/else just like with `g`, but doing that just shifts the error to something else in DynamicPPL, I think `from_maybe_linked_vec_transform`.

mhauru:
Attempting to fix #207. The current version fixes the MWE, but I've failed to make it into a test (the MWE stops crashing once you wrap it in a testset). Hoping to run the Turing tests against this though.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,7 @@
+# 0.9.10
+
+Fix a bug introduced in 0.9.9 that made certain phi nodes with Union types fail a type assertion.
+
 # 0.9.9
 
 Remove manual opaque closure optimisation functions in favour of setting the world age and letting the compiler do more work for us, and providing it with some more type information. This changes no functionality, and shouldn't change performance either, but simplifies code.
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "6f1fad26-d15e-5dc8-ae53-837a1d7b8c9f"
 license = "MIT"
 desc = "Tape based task copying in Turing"
 repo = "https://github.com/TuringLang/Libtask.jl.git"
-version = "0.9.9"
+version = "0.9.10"
 
 [deps]
 MistyClosures = "dbe65cb8-6be2-42dd-bbc5-4196aaced4f4"


mhauru:
Local testing indicates that the particle MCMC tests in Turing should pass with this (at least most do), and the performance gains are kept. Still need to figure out how to do the test case for this in Libtask's suite.

mhauru:
The integration tests are still running, but I did the crucial parts locally and so far they seem fine on CI too, and I need to run off now until Monday, so I'm just gonna risk it. It went so well with #205.

PR #213, opened 2026-02-21
Clean up benchmarks folder and CI

penelopeysm:
This PR gets rid of the severely outdated scripts in the `perf` folder (nobody bothered to update them, even when Libtask got rewritten, and even when there was a major effort in updating Libtask for Julia 1.12). 

In its place we add a new set of integration tests with Turing. This set is fairly basic and just aims to show that sampling with SMC / PG actually works. This replaces the old strategy of running the entire Turing test suite, which was wasteful and led to incidents like #207.

Over time this set of tests can be expanded to cover other problematic models, such as those in the other open issues.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -8,8 +8,8 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
-    fail-fast: false
     strategy:
+      fail-fast: false
       matrix:
         version:
           - 'min'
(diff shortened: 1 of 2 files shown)


penelopeysm:
We are spawning a slightly silly number of runners for the new Turing integration test. I think it is important to make sure that the Turing integration test passes on all combinations, because Turing **is** the raison d'etre of Libtask. We should test Turing integration as seriously as we test Libtask itself.

